PR #1, opened 2016-04-28
Made m_wrapper public to be able to compile artery

JohanBillman:
gcc 5.3.1 failed compilation due to m_wrapper being private


riebl:
This error was caused by a mismatch of the used Artery and Vanetza versions. Fixed by artery/8a41461.

PR #4, opened 2016-08-14
H_Da Student : Socktap for Vanetza

sachinkashyapbs:
This implementation contains :
-> Logic to open the file contianing GPS data and reading NMEA RMC
sentences.
-> Processing these sentences and obtaininig the data in string format,
which are needed to update the long position vector object.


riebl:
See c439021dc91a378902fae2ffe9dd6fda62227200
Some features, e.g. providing GNSS positions from trace files, are postponed for future inclusion.

PR #7, opened 2017-10-22
Allow usage of GPSD faker

kelunik:
I currently use https://github.com/loewexy/gpsd-fake and commented out the same address check to send and receive messages on the local loopback inteface `lo`.

https://github.com/riebl/vanetza/blob/7548565d39d76bc8a5015d029edc5cf83c94e101/vanetza/geonet/router.cpp#L873

The above check fails if packets aren't skipped as long as a valid GPS position isn't available yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+build
 conanfile.pyc
 doc
--- a/tools/socktap/CMakeLists.txt
+++ b/tools/socktap/CMakeLists.txt
@@ -8,6 +8,7 @@ find_package(GPS REQUIRED)
 
 add_executable(socktap
     application.cpp
+    dcc_passthrough.cpp
     ethernet_device.cpp
     gps_position_provider.cpp
     hello_application.cpp
--- a/tools/socktap/dcc_passthrough.cpp
+++ b/tools/socktap/dcc_passthrough.cpp
@@ -1,3 +1,4 @@
+#include "dcc_passthrough.hpp"
 #include "time_trigger.hpp"
 #include <vanetza/dcc/data_request.hpp>
 #include <vanetza/dcc/interface.hpp>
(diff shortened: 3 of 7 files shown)


riebl:
Hm, there is something odd: `is_empty(source_entry.position_vector)` is false if a non-zero GeoNet address is used and the address does not depend on GPS availability. However, Linux assigns a zero MAC address to `lo`. I suppose generating a more sophisticated GeoNet address in `RouterContext` would fix the issue, e.g. by adding `mib.itsGnLocalGnAddr.is_manually_configured(true);` to https://github.com/riebl/vanetza/blob/master/tools/socktap/router_context.cpp#L50

Could you please verify that this fixes the reported issue?

kelunik:
It does indeed assign a all-zero MAC address, but why does that matter? With the current fix it works totally fine for me after a few seconds. Your suggested fix seems to work as well.

`is_empty(source_entry.position_vector)` checks the GPS position vector, no? I can't see how that's related to the MAC address? What am I missing?

riebl:
As defined by ETSI EN 302 636-4-1, GeoNet position vectors always contain the address and timestamp belonging to the position coordinates. `is_empty` checks all attributes of the position vector not only the coordinates. Hence, the MAC address matters in this case :-)

One can argue of course if packet generation shall be suppressed until a position fix is available. However, I don't like your approach using a global variable. Alternatively, you could toggle the behaviour at `RouterContext::update_position_vector` invocations by a new `DccPassthrouh::allow_packet_flow(bool)` method.

kelunik:
Should that new method be added to the `RequestInterface` or rather just the type in the `RouterContext` changed?

riebl:
I would change the type of `RouterContext`'s request_interface_ and keep the `RequestInterface` lean. Maybe it is about time to add some more runtime options to socktap (e.g. for configuring the gpsd endpoint [memory vs. socket] or even providing a fixed predefined position without gpsd).

kelunik:
@riebl I updated the PR and moved `DssPassthrough` into its own file. I plan to add configuration options in the near future, because I want to have a simple override for the MAC address anyway, so I can simply run two instances on the localhost, which don't share the same address.

Currently the build fails with the following error, I'm probably missing something really simple, but no idea what.

```
[ 98%] Linking CXX executable ../../bin/socktap
CMakeFiles/socktap.dir/router_context.cpp.o: In function `RouterContext::RouterContext(boost::asio::basic_raw_socket<boost::asio::generic::raw_protocol, boost::asio::raw_socket_service<boost::asio::generic::raw_protocol> >&, EthernetDevice const&, TimeTrigger&, PositionProvider&)':
router_context.cpp:(.text+0x1dc): undefined reference to `DccPassthrough::DccPassthrough(boost::asio::basic_raw_socket<boost::asio::generic::raw_protocol, boost::asio::raw_socket_service<boost::asio::generic::raw_protocol> >&, TimeTrigger&)'
CMakeFiles/socktap.dir/router_context.cpp.o: In function `RouterContext::update_position_vector()':
router_context.cpp:(.text+0xccd): undefined reference to `DccPassthrough::allow_packet_flow(bool)'
CMakeFiles/socktap.dir/router_context.cpp.o: In function `DccPassthrough::~DccPassthrough()':
router_context.cpp:(.text._ZN14DccPassthroughD2Ev[_ZN14DccPassthroughD5Ev]+0xf): undefined reference to `vtable for DccPassthrough'
collect2: error: ld returned 1 exit status
tools/socktap/CMakeFiles/socktap.dir/build.make:291: recipe for target 'bin/socktap' failed
make[2]: *** [bin/socktap] Error 1
CMakeFiles/Makefile2:727: recipe for target 'tools/socktap/CMakeFiles/socktap.dir/all' failed
make[1]: *** [tools/socktap/CMakeFiles/socktap.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

riebl:
Ensure that DccPassthrough.cpp (or whatever the name of the newly created source file for `DccPassthrough` is) is listed in tools/socktap/CMakeLists.txt along with the other sources. This should fix the above issue.

riebl:
I suppose this is caused by the `using namespace`stuff in dcc_passthrough.hpp. As a rule of thumb, there should be no `using ...` or `namespace = ...` in header files (this is not specific to Vanetza).

kelunik:
Removing the `using` statements doesn't change anything, but I fixed it now. :man_facepalming:

kelunik:
Closing, as it has been merged / is contained in #8.

kelunik:
@riebl Regarding the failing `assert()` without `is_manually_configured(true)`: I guess the remote side shouldn't be able to trigger assertion failures that crash the program.

riebl:
@kelunik Yes, I agree with you in principle. If there is a bug, however, I would still allow incoming packets to crash the program: Fixing bugs is the right way to increase robustness (not removing assertions) :-)

kelunik:
@riebl Sure, but it's a bug in the remote, not the local program in that case, no?

riebl:
cd8d95707b82c26386a96a80b2210accee2ae72e fixes the assertion being triggered by malicious incoming packets with position vectors fulfilling the `is_empty` test.
@kelunik At least it is unexpected behaviour by the remote side. The assertion is only there to catch potential LocationTable bugs (returning empty entry though LocationTable::update has been called right before).

kelunik:
Did you link the wrong commit?

kelunik:
Yes, c6918af9659d50ed75bca62eb16788a25e8349c1 is the right one.

PR #8, opened 2017-11-09
Add runtime options to socktap

kelunik:
This builds on #7.

This adds runtime options to socktap and includes a mechanism to fake the sender MAC address. It allows running socktap on the loopback interface with two different addresses easily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+build
 conanfile.pyc
 doc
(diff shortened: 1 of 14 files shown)


riebl:
389e4cb includes all the features requested by you (with some changes). Thank you for contributing to Vanetza!

kelunik:
Yes, I have seen that. The `FakeNetworkDevice` is indeed unnecessary. I have a couple of other commits locally right now, I'll open a PR for those as soon as I've rebased them onto the current `master`.

PR #10, opened 2017-11-13
socktap: Implement optional security

kelunik:
`--security` can be set to either `'off'` or `'naive'` to sign messages. Signed messages are currently not validated, but always pass as OK.

### DIFF
--- a/tools/socktap/main.cpp
+++ b/tools/socktap/main.cpp
@@ -8,6 +8,8 @@
 #include <boost/asio/generic/raw_protocol.hpp>
 #include <boost/program_options.hpp>
 #include <iostream>
+#include <vanetza/security/naive_certificate_manager.hpp>
+#include <vanetza/security/security_entity.hpp>
 
 namespace asio = boost::asio;
 namespace gn = vanetza::geonet;
@@ -21,6 +23,7 @@ int main(int argc, const char** argv)
         ("help", "Print out available options.")
         ("interface,i", po::value<std::string>()->default_value("lo"), "Network interface to use.")
         ("mac-address", po::value<std::string>(), "Override the network interface's MAC address.")
+        ("security", po::value<std::string>()->default_value("off"), "Security profile to use.")
         ("gpsd-host", po::value<std::string>()->default_value(gpsd::shared_memory), "gpsd's server hostname")
         ("gpsd-port", po::value<std::string>()->default_value(gpsd::default_port), "gpsd's listening port")
         ("require-gnss-fix", "suppress transmissions while GNSS position fix is missing")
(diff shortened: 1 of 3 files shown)


tools/socktap/main.cpp
```diff
@@ -85,11 +88,36 @@ int main(int argc, const char** argv)
...
+
+        std::string securitySetting = vm["security"].as<std::string>();
+
+        // We always use the same ceritificate manager and crypto services for now.
+        // If itsGnSecurity is false, no signing will be performed, but receiving of signed messages works as expected.
+        auto certificate_manager_factory = security::builtin_certificate_managers();
+        auto certificate_manager = certificate_manager_factory.create("Naive", trigger.runtime());
+
+        auto crypto_backend_factory = security::builtin_backends();
+        auto crypto_backend = crypto_backend_factory.create();
```
riebl: You can simplify these two lines:
`auto crypto_backend = security::create_backend("default");`

tools/socktap/router_context.cpp
```diff
@@ -24,10 +24,10 @@ RouterContext::RouterContext(raw_protocol::socket& socket, const geonet::MIB& mi
     router_.set_address(mib_.itsGnLocalGnAddr);
     router_.set_access_interface(request_interface_.get());
     router_.set_transport_handler(geonet::UpperProtocol::BTP_B, &dispatcher_);
-    update_position_vector();
+    router_.set_security_entity(&security_entity);
 
+    update_position_vector();
     do_receive();
-    trigger_.schedule();
```
riebl: Why do you remove the scheduling of the time trigger?

tools/socktap/main.cpp
kelunik: Done. :+1:

tools/socktap/router_context.cpp
kelunik: I initially removed it, because I introduced a setter for the security entity. If the security entity isn't set, but `MIB::itsGnSecurity` is true, that one causes an error because of an unset security entity, which I solved by removing it and deferring the `update_position_vector()` by 0 seconds, which allows setting the security entity right after the constructor call.

In the squashed commit I went with setting the security entity always (hence it's in the constructor), as otherwise no signed packets can be received.

I left it removed, because it's anyway directly called in `update_position_vector()`.

I can of course move that to a separate commit or remove it entirely from the commit.

riebl: Okay, I see. Removed it from your commit.

riebl:
Merged 1cf37bf3251f0d0b984ae0568d7098566945b600. Thank you!

PR #11, opened 2017-11-13
Reject certificates without supported time constraint

kelunik:
Certificates should always have a time constraint. Currently only `time_start_and_end` is supported. If no such constraint is given, certificates were not rejected before. That means any certificate using `time_end` or `time_start_and_duration` was simply deemed valid.


riebl:
Merged f73892879580302dc16d81f858de4017745268b3

PR #12, opened 2017-11-15
Add support for time_end in NaiveCertificateManager

kelunik:
I can also add support for `time_start_and_duration`, but I'm not sure which unit to use, as the unit for `Duration` can be years.

### DIFF
--- a/vanetza/security/validity_restriction.hpp
+++ b/vanetza/security/validity_restriction.hpp
@@ -5,6 +5,7 @@
 #include <vanetza/security/region.hpp>
 #include <vanetza/security/serialization.hpp>
 #include <boost/variant/variant.hpp>
+#include <chrono>
 
 namespace vanetza
 {
@@ -34,6 +35,24 @@ public:
     {
         return m_raw;
     }
+
+    /**
+     * Returns the duration unit.
+     */
+    Units unit() const
+    {
+        return static_cast<Units>(m_raw >> 13);
+    }
+
+    uint16_t value() const
+    {
+        return m_raw & 0xCFFF; // mask upper 3 bit
+    }
+
+    /**
+     * Returns the duration in seconds.
+     */
+    std::chrono::seconds to_seconds() const;
 
 private:
     uint16_t m_raw;
@@ -108,4 +127,3 @@ void serialize(OutputArchive&, const ValidityRestriction&);
 } // namespace vanetza
 
 #endif /* VALIDITY_RESTRICTION_HPP_LMCUHYLJ */
-
--- a/vanetza/security/validity_restriction.hpp
+++ b/vanetza/security/validity_restriction.hpp
@@ -46,7 +46,7 @@ public:
 
     uint16_t value() const
     {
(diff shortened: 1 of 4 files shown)


riebl:
I see no problem to use `vanetza::common::Clock` which stores microseconds in a 64bit integer (could enough for > 200.000 years). A conversion method from `security::Duration` to `std::chrono::duration` should be easy to implement.
Thanks for this pull request!

kelunik:
I simply went with `Time64` now, which is already defined inside vanetza.

vanetza/security/validity_restriction.hpp
```diff
@@ -35,6 +36,47 @@ class Duration
...
+
+    uint16_t value() const
+    {
+        return m_raw & 0xCFFFF; // mask upper 3 bit
+    }
+
+    /**
+     * Returns the duration in seconds.
+     */
+    Time64 to_seconds() const
```
riebl: This is not possible: `Time64` encodes microseconds (see section 4.2.15 of TS 103 097).
Exactly because such conversion errors happen too easily with plain integers I would favor a solution based on `vanetza::Clock::duration`.

vanetza/security/validity_restriction.hpp
```diff
@@ -35,6 +36,47 @@ class Duration
...
+                return value() * 216000;
+
+            case Units::Years:
+                return value() * 31556925;
+
+            default:
+                // undefined, let's interpret it as a duration of 0, which will fail during validation
+                return 0;
+        }
+    }
```
riebl: Larger code segments should go to the source file, i.e. everything except simple getter/setter and template code.

vanetza/security/validity_restriction.hpp
```diff
@@ -35,6 +36,47 @@ class Duration
...
+     * Returns the duration unit.
+     */
+    Units unit() const
+    {
+        return static_cast<Units>(m_raw >> 13);
+    }
+
+    uint16_t value() const
+    {
+        return m_raw & 0xCFFFF; // mask upper 3 bit
```
riebl: mask has to be 0xCFFF

kelunik:
I've fixed the mentioned points and added a few unit tests.

Currently `make` seems to recompile all tests every time, is there any way to only recompile the changed files and the dependent files instead?

riebl:
Is it really recompiling all tests? On my system it does only re-link those unit tests depending on the security module (e.g. the geonet tests).

kelunik:
Seems like I just got the impression because I changed the mask in `validity_restriction.hpp` which compiled a very large set of tests.

riebl:
Merged 0a87827bc7ce019955fb45e7f9e151b9d9ab2599

kelunik:
Does `to_seconds` still make sense with `std::chrono::duration`?

riebl:
`security::Duration::to_seconds` returns `std::chrono::seconds`, so this makes sense to me. If it were returning `Clock::duration` I would probably change it to a cast operator. Since there are no rounding errors when converting `security::Duration` to `Clock::duration` this does not introduce problems from my point of view.

PR #14, opened 2017-11-18
socktap: Print human readable drop reason

kelunik:
Otherwise one has to search through the code to find out the drop reason.


riebl:
Merged 09d14f22842e56e3142f3ca97b45ceb8b468ffe9

PR #16, opened 2017-11-18
Add 'tags' to .gitignore

kelunik:
File used by Atom for autocomplete powered by ctags.


riebl:
In my opinion it's not a good idea to have "personal ignore patterns" in .gitignore. Please use .git/info/exclude for this purpose. If you think there is still a valid reason for adding this to .gitignore then please re-open this ticket.

PR #17, opened 2017-11-18
certify: New command line tool for certificates

kelunik:
Certify allows creating keys and certificates from the command line for later use in applications.

A follow-up commit will add support for these certificates in a new certificate manager.

This is an initial version. Command names and options can probably be improved as well as some parts of the certification process can be extracted info functions.

I'm creating this PR, so comments can be added.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(BUILD_SOCKTAP)
 endif()
 
 option(BUILD_CERTIFY "Build certify application" OFF)
-if(BUILD_SOCKTAP)
+if(BUILD_CERTIFY)
   add_subdirectory(tools/certify)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,11 @@ if(BUILD_SOCKTAP)
   add_subdirectory(tools/socktap)
 endif()
 
+option(BUILD_CERTIFY "Build certify application" OFF)
+if(BUILD_SOCKTAP)
+  add_subdirectory(tools/certify)
+endif()
+
 # interface library for convenience
 get_property(_components GLOBAL PROPERTY VANETZA_COMPONENTS)
 add_library(vanetza INTERFACE)
@@ -110,4 +115,3 @@ file(COPY ${CMAKECONFIG_BUILD_DIR}/VanetzaConfigVersion.cmake
 configure_file(cmake/VanetzaExportsConfig.cmake.in
   ${CMAKE_CURRENT_BINARY_DIR}/VanetzaConfig.cmake @ONLY)
 export(PACKAGE ${PROJECT_NAME})
-
--- a/tools/certify/command.hpp
+++ b/tools/certify/command.hpp
@@ -1,0 +1,13 @@
+#ifndef CERTIFY_COMMAND_HPP
+#define CERTIFY_COMMAND_HPP
+
+#include <boost/program_options.hpp>
+
+class Command
+{
+public:
+    virtual int execute() = 0;
+    virtual void parse(std::vector<std::string>&) = 0;
+};
+
+#endif /* CERTIFY_COMMAND_HPP */
--- a/tools/certify/command.hpp
+++ b/tools/certify/command.hpp
@@ -1,7 +1,8 @@
 #ifndef CERTIFY_COMMAND_HPP
 #define CERTIFY_COMMAND_HPP
 
-#include <boost/program_options.hpp>
+#include <string>
+#include <vector>
 
 class Command
 {
--- a/tools/certify/commands/auth-ca.cpp
+++ b/tools/certify/commands/auth-ca.cpp
@@ -1,6 +1,7 @@
-#include "../keyio.hpp"
 #include "auth-ca.hpp"
+#include "keyio.hpp"
 #include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
 #include <chrono>
 #include <cryptopp/eccrypto.h>
 #include <cryptopp/oids.h>
--- a/tools/certify/commands/auth-ca.cpp
+++ b/tools/certify/commands/auth-ca.cpp
@@ -1,0 +1,127 @@
+#include "../keyio.hpp"
+#include "auth-ca.hpp"
+#include <boost/filesystem.hpp>
+#include <chrono>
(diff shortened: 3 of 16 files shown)


CMakeLists.txt
```diff
@@ -66,6 +66,11 @@ if(BUILD_SOCKTAP)
   add_subdirectory(tools/socktap)
 endif()
 
+option(BUILD_CERTIFY "Build certify application" OFF)
+if(BUILD_SOCKTAP)
```
riebl: typo: BUILD_CERTIFY

tools/certify/command.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef CERTIFY_COMMAND_HPP
+#define CERTIFY_COMMAND_HPP
+
+#include <boost/program_options.hpp>
```
riebl: unused include, however, vector and string includes are missing

tools/certify/commands/auth-ca.cpp
```diff
@@ -0,0 +1,127 @@
+#include "../keyio.hpp"
```
riebl: This looks clumsy. Use INCLUDE_DIRECTORIES property in CMake if necessary to get rid of relative paths here.

tools/certify/options.hpp
```diff
@@ -0,0 +1,10 @@
+#ifndef CERTIFY_OPTIONS_HPP
+#define CERTIFY_OPTIONS_HPP
+
+#include "command.hpp"
+#include <boost/variant/variant.hpp>
+#include <boost/variant/get.hpp>
```
riebl: boost::variant is not used here

tools/certify/commands/auth-ca.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+
+class AuthCaCommand : public Command {
+public:
+    std::string output;
+    std::string sign_key;
+    std::string sign_cert;
+    std::string subject_key;
+    std::string subject_name;
+    int validity_days;
```
riebl: Avoid public attributes whenever possible (and they seem to be unnecessary for `Command`).

kelunik: Indeed, I previously parsed the arguments inside `options.cpp` and moved it into the commands now.

tools/certify/commands/auth-ca.cpp
kelunik: I'll have a look, I'm still new to CMake.

tools/certify/options.hpp
kelunik: Leftover, will update.

kelunik:
I have fixed the mentioned points in a new commit instead of modifying the previous commit, so you can see the changes. I can squash them before merging, but I guess you'll do that anyway, because you merge PRs manually.

kelunik:
@riebl Do you know whether there's a certificate signing request (CSR) definition yet? IEEE 1609.2-2016 seems to have a glossary definition of CSR, but doesn't actually use it anywhere in the document.

tools/certify/keyio.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+    vanetza::security::BackendCryptoPP::PublicKey public_key;
+    private_key.MakePublicKey(public_key);
+
+    auto& public_element = public_key.GetPublicElement();
+    public_element.x.Encode(kp.public_key.x.data(), kp.public_key.x.size());
+    public_element.y.Encode(kp.public_key.y.data(), kp.public_key.y.size());
+
+    return kp;
+}
```
kelunik: @riebl Do you think it makes sense to make these functions from `BackendCryptoPP` public instead?

```cpp
/// create private key
PrivateKey generate_private_key();

/// derive public key from private key
PublicKey generate_public_key(const PrivateKey&);

/// adapt generic public key to internal structure
PublicKey internal_public_key(const ecdsa256::PublicKey&);

/// adapt generic private key to internal structure
PrivateKey internal_private_key(const ecdsa256::PrivateKey&);
```

These would also be usable in `genkey` then.

riebl:
@kelunik there is an "Opaque" signedCertificateRequest in Ieee1609dot2Content (ASN.1) that I am aware of. IEEE 1609.2 says: "signedCertificateRequest indicates that the content is a certificate request. Further specification of certificate requests is not provided in this version [2016] of this standard." I have no idea why it is included at all then...

tools/certify/keyio.cpp
riebl: @kelunik You seem to have found a good reason to make these methods public. At a first glance, these methods won't break any `BackendCryptoPP` invariant because `m_prng` is the only used member attribute.

kelunik:
I've also found http://oldforge.etsi.org/websvn/listing.php?repname=ITS.ITS&path=%2Ftrunk%2Ftools%2F&#abf717a9dc39e80779f00e1ec07bb728e, which offers a cert generation tool and checker, but which doesn't even compile fine for me without changes. Currently it shows the generated certificates as self-signed, even if they shouldn't be self-signed. I'll investigate.

kelunik:
I have found [ETSI TS 102 941 v1.1.1](http://www.etsi.org/deliver/etsi_ts/102900_102999/102941/01.01.01_60/ts_102941v010101p.pdf) which specifies the requests. It's quite strange that this builds on IEEE 1609.2 while published in 2012, ETSI then built the other security features on its own (I think mostly with different naming from IEEE 1609.2), and with the 2017 version they went with IEEE 1609.2 again.

While an implementation with CSRs is definitely required for production use, I think an implementation with public keys as input is totally fine for testing purposes. Currently my code only supports loading private keys, but that will change, as it's not possible to issue certificates for systems using HSMs otherwise.

riebl:
@kelunik Is this a work-in-progress PR or shall it be merged already?

kelunik:
This PR can be merged, I can add a command to extract the public key from a private key later. It already works fine for testing.

riebl:
Merged with 8b7a82a9ae247185e054265637b584d7d639f27d.
I have made some minor modifications in the follow-up commit 9b2d9c21eed3e1c79b7a1b4a7755cf0dfa34302b. Sorry for the late merge, I am currently struggling with some hardware issues...

kelunik:
Don't worry, hope your hardware issues are solved soon. :)

PR #20, opened 2017-11-20
socktap: Separate certificate validation and certificate / key provisioning

kelunik:
This PR separates certificate validation and certificate / key provisioning. This avoids coping the `check_certificate` of `NaiveCertificateManager` to other certificate managers, just because the provisioning is different.

This PR also introduces a `TrustStore` that allows trusting a number of signers instead of using a magic hash. Chain validation will be provided in a different PR.


vanetza/security/naive_certificate_provider.cpp
```diff
@@ -0,0 +1,142 @@
...
+    certificate.subject_info.subject_name = subject;
+
+    // section 6.6 in TS 103 097 v1.2.1 - levels currently undefined
+    certificate.subject_attributes.push_back(SubjectAssurance(0x00));
+
+    // section 7.4.1 in TS 103 097 v1.2.1
+    // set subject attributes
+    // set the verification_key
+    Uncompressed coordinates;
+    coordinates.x.assign(m_root_key_pair.public_key.x.begin(), m_root_key_pair.public_key.x.end());
```
kelunik: @riebl This currently segfaults in several unit tests, but I fail to see why for quite some time now. Could you have a look?

```
#11 0x00005555559b05f4 in std::vector<unsigned char, std::allocator<unsigned char> >::assign<unsigned char const*, void> (this=0x7fffffffd4d0, 
    __first=0x20 <error: Cannot access memory at address 0x20>, 
    __last=0x40 <error: Cannot access memory at address 0x40>)
    at /usr/include/c++/7/bits/stl_vector.h:521

#12 0x00005555559afbeb in vanetza::security::NaiveCertificateProvider::generate_root_certificate (this=0x555555d4c0d0)
    at /home/kelunik/GitHub/riebl/vanetza/vanetza/security/naive_certificate_provider.cpp:112
```

If I try to avoid the reference for the root pair and use a freshly generated pair like for the `own_key_pair`, then it fails with that error instead:

```
C++ exception with description "CryptoMaterial: this object contains invalid values" thrown in the test fixture's constructor.
```

riebl: I think I have found the reason for your segfaults, see my branch kelunik-basic-cert-mngr at 8709f1ed982f53815abbf0d47e7fd4f9f09ed83e.
However, some unit tests still fail (no segfault).

kelunik: Thanks, I've just pushed a commit that fixes those failures as well. It just used the wrong certificate as signer.

riebl:
@kelunik is this pull request still work in progress?

kelunik:
@riebl Yes, there's still a small change I'd like to make: Incorporate the `TrustStore` into the `DefaultCertificateValidator`.

riebl:
@kelunik Alright, just ping me when I shall have a look :-)

kelunik:
@riebl Will do. Currently I have the same issue with the root cert as previously with the root key. Why does it not work with storing a reference? The memory of a `static ...` shouldn't be freed until the program ends.

riebl:
@kelunik Are you sure that the reference to a `static` causes the problem? Previously, `std::vector<unsigned char> subject(m_root_certificate_subject.c_str(), m_root_certificate_subject.c_str() + m_root_certificate_subject.size() + 1);` was causing trouble. Do you have something like that in your root certificate related code?

kelunik:
I thought members would be initialized based on the order in the initializer list, but TIL they're initialized based on the declaration order.

I'm fine with merging now if you're fine, too. I'll add further features in future PRs. Should I change `INVALID_ROOT_HASH` to `UNKNOWN_SIGNER`? We can also delete `INVALID_SIGNATURE` now, as it will never be returned. Even if the signature is wrong, it might be a colliding `HashedId8` and the certificate just unknown.

vanetza/security/default_certificate_validator.cpp
```diff
@@ -0,0 +1,122 @@
...
+        return CertificateInvalidReason::INVALID_NAME;
+    }
+
+    // check signer info
+    if (get_type(certificate.signer_info) != SignerInfoType::Certificate_Digest_With_SHA256) {
+        return CertificateInvalidReason::INVALID_ROOT_HASH;
+    }
+
+    HashedId8 signer_hash = boost::get<HashedId8>(certificate.signer_info);
+    std::vector<Certificate> possible_signers = trust_store.find_by_id(signer_hash);
```
namnatulco: Wouldn't it be better to fetch the claimed signer from the certificate and then verify whether that signer is actually valid? If I understand this part correctly, you're taking the signers from the trust store and checking whether you have the signer in your trust store. I would have expected that you fail here if the returned vector does not have length 1 (reason: untrusted cert). If this list has more than one entry, you've got a collision in your hash, and that should never happen, right?

vanetza/security/default_certificate_validator.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (!sig) {
+        return CertificateInvalidReason::MISSING_SIGNATURE;
+    }
+
+    // create buffer of certificate
+    ByteBuffer cert = convert_for_signing(certificate);
+
+    bool valid_signature = false;
+
+    for (Certificate& possible_signer : possible_signers) {
```
namnatulco: if you do that, this iteration would always happen exactly once (i.e., you'd have exactly one verification call to the crypto backend).

vanetza/security/default_certificate_validator.cpp
namnatulco: Separate from that, I don't really understand how you're dealing with chains here. As far as I can tell this code assumes that the signer of the certificate currently being checked is already in the trust store (i.e., it's a root certificate). I don't think it is a good idea to cache intermediate CA certs in the trust store, unless you have some kind of periodic removal in place. But maybe I just misunderstood your code here.

kelunik: A collision in the hash can always happen, it's not that likely but might happen. A `HashedId8` isn't a full SHA-256 hash but only the first 8 bytes (=64 bit) of it. Even with a full hash it might happen theoretically.

You're right that this currently requires the signer to be directly in the trust store. Full chain validation will follow in another PR.

vanetza/security/default_certificate_validator.cpp
kelunik: If there's no collision, right. If there's no certificate found, it will just fail with `UNKNOWN_SIGNER`.

vanetza/security/default_certificate_validator.cpp
namnatulco: The problem is that if it happens, you can't really trust the result anymore either, which is why I suggested length != 1. It has the (very minor) added advantage that the validation time will be more consistent.

kelunik: Yes, correct. Chain validation will follow in a separate PR and will likely require changes here. It's mostly required for the separation change that doesn't rely on a magic hash anymore.

namnatulco: 👍 awesome. It might be nice to have separate stores for the trusted certs and the "validation cache" (i.e., where previously validated intermediate certs, or even vehicle certs, can be cached).

kelunik: Yes, it needs some additional API changes to continue the chain checking with the right one in that case.

kelunik: Yes, these will be separate, as root certs won't expire. But there might be an additional entity that combines both to make lookups easier.

riebl:
@namnatulco @kelunik Could you please open issues concerning the remaining security issues (e.g. chain validation) so the outcome of your discussion here is documented? I am not sure if I have understood all the subtle details.

kelunik:
@riebl I've opened #23.

riebl:
Changes merged with 0024927891dae9d508df498d1ddb0c33224bc7e4

@kelunik Thank you. Do you have plans to tidy up socktap/main.cpp? This file became a little bit bloated recently.

kelunik:
@riebl Didn't have plans, yet, but I can probably do that in the coming days. One thing that needs to happen anyway is to get rid of the hardcoded filepaths I left in there, oops.

riebl:
@kelunik I would appreciate moving the key file handling to vanetza::security so certify and socktap can both use it. I have created #24 for this.

PR #21, opened 2017-11-20
socktap: Prevent stale clock during security validation

### LINKED ISSUE #19: check_generation_time too strict
It seems like [`check_generation_time`](https://github.com/riebl/vanetza/blob/616fb558dd297fc74886e117c7592e19833bc958/vanetza/security/verify_service.cpp#L16-L46) is too strict.

https://github.com/riebl/vanetza/blob/616fb558dd297fc74886e117c7592e19833bc958/vanetza/security/verify_service.cpp#L25

That requires clocks to be out-of-sync no more than 40 milliseconds, otherwise that check fails.

Currently I hit that even locally with the same clock, but because of a different issue: the clock is currently only updated in each `Runtime` tick.

kelunik:
Fixes #19.


kelunik:
Closing, as merged via https://github.com/riebl/vanetza/issues/19#issuecomment-345744235

PR #27, opened 2017-12-09
Add basic chain validation

kelunik:
Support for caching chains has been added as well as validating chains constructed from cached entries.

There are currently no consistency checks implemented. Sending unrecognized certificate requests also isn't implemented, yet.

As part of the proper chain validation a few tests had to be adjusted. These relied on certificate validation happening before the message itself is checked. These use the modified certificate when signing now, so these messages don't end up returning a false signature error only.

### DIFF
--- a/tools/socktap/hello_application.cpp
+++ b/tools/socktap/hello_application.cpp
@@ -22,7 +22,7 @@ void HelloApplication::indicate(const DataIndication& indication, UpPacketPtr pa
 
 void HelloApplication::schedule_timer()
 {
-    timer_.expires_from_now(std::chrono::seconds(2));
+    timer_.expires_from_now(std::chrono::milliseconds(800));
     timer_.async_wait(std::bind(&HelloApplication::on_timer, this, std::placeholders::_1));
 }
 
--- a/tools/socktap/main.cpp
+++ b/tools/socktap/main.cpp
@@ -171,7 +171,8 @@ int main(int argc, const char** argv)
             mib.itsGnSecurity = false;
         }
 
-        security::SignService sign_service = straight_sign_service(trigger.runtime(), *certificate_provider, *crypto_backend);
+        security::SignPreparer sign_preparer(trigger.runtime().now());
+        security::SignService sign_service = straight_sign_service(trigger.runtime(), *certificate_provider, *crypto_backend, sign_preparer);
         security::VerifyService verify_service = straight_verify_service(trigger.runtime(), *certificate_validator, *crypto_backend, cert_cache);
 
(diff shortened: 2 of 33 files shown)


kelunik:
@riebl Does it make sense to drop `vanetza::security::Profile`? The `SignRequest` already contains the `its_aid` which can equally be used and is currently used in my implementation.

Currently all messages are signed as with the CAM profile due to:

https://github.com/riebl/vanetza/blob/63d884992664f720a5533f7b052699249c9c3dc4/vanetza/security/security_entity.cpp#L29

I also think it makes sense turning `SignService` into an object instead of it using the currently implemented `SignPreparer` (I didn't have a better name). We could override `operator()` to keep the BC impact low. What's your opinion there?

Currently there's a failing test, because the repetition packets aren't the same size as the first packet anymore in `RouterRequest`. I'm not sure how to fix that without using a fixed offset.

riebl:
@kelunik Yes, from my point of view `vanetza::security::Profile` can be deprecated. It has been part of the security interface in earlier ETSI documents but seems to be replaced by ITS AIDs everywhere.
I will have a closer look at your PR as soon as possible.

riebl:
@kelunik I just had a look at `SignPreparer`: Maybe `SignHeaderPolicy` could be a fitting name for its purpose. However, `std::list<HeaderField> SignHeaderPolicy::prepare_header(const SignRequest&, const Certificate&)` should be a sufficient interface, isn't it? I propose to pass the `Runtime` object to SignHeaderPolicyImpl's constructor instead of passing it a each prepare_header call. 
This way we don't end up with a single class doing everything and it should be fairly easy to implement different header policies.

kelunik:
@riebl That sounds like a very good name. Apart from the policies there needs to be an entity that decides which policy to use, as a single security entity might handle messages from multiple applications.

I'd be glad if you could have a look at the failing test and propose a good way to fix it, the review of the rest of the PR can wait, as there will probably be further changes.

> Currently there's a failing test, because the repetition packets aren't the same size as the first packet anymore in `RouterRequest`. I'm not sure how to fix that without using a fixed offset.

riebl:
@kelunik: I had to fix a compiler issue in SecurityEntityTest and relaxed the failing RouterRequest test case a little bit. This test case is probably older than the security implementation ;-)
See my cert-chains branch.

riebl:
@kelunik Is this PR ready for approval now?

kelunik:
Yes, it is. I'll add revocation checks and stricter certificate checks in a follow-up PR.

riebl:
Finally merged b8c601ce5782f456b11c570cd483d52e888d521e

PR #29, opened 2018-01-06
common: Add parsing support for BTP-A

kelunik:
I noticed these aren't supported while trying to implement an upper tester for the TTCN-3 test suite for BTP.


riebl:
Merged 544f2a197d86adc36e9757a84a127931d34e3b82, thanks!

PR #30, opened 2018-01-11
Upper Tester

kelunik:
This implements a basic upper tester for ETSI's TTCN-3 test suite. It currently only implements the necessary payloads for the BTP tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,11 @@ if(BUILD_CERTIFY)
   add_subdirectory(tools/certify)
 endif()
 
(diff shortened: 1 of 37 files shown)


kelunik:
@riebl Should I make Travis run these tests as part of the CI?

riebl:
@kelunik That would be awesome! I am not sure if it is possible to include sub-scripts to avoid having a huge single Travis recipe?

kelunik:
@riebl What do I need to change to make https://travis-ci.org/riebl/vanetza/builds/337304318#L4015 work? Works fine for me locally. Guess it has to do with different versions.

PR #34, opened 2018-01-28
Certificate validator improvements

kelunik:
It now checks the validity restrictions of certificates in the trust store and applies rectangular and circular region restrictions. It fails closed if any other (currently unsupported) region restrictions are in place.

Additionally, this adds a couple of tests, most of which are ported from the ETSI test suite, which I'm not yet able to run.

### DIFF
--- a/vanetza/security/default_certificate_validator.cpp
+++ b/vanetza/security/default_certificate_validator.cpp
@@ -114,8 +114,11 @@ void DefaultCertificateValidator::update(const LongPositionVector& lpv)
     m_local_position_vector.gn_addr = gn_addr;
 }
 
-CertificateValidity DefaultCertificateValidator::check_certificate(const Certificate& certificate)
+DecapConfirm DefaultCertificateValidator::check_certificate(const Certificate& certificate)
 {
+    DecapConfirm confirm;
+    std::list<Certificate> in_progress_chain; // in progress chain, don't assign to confirm unless complete
+
     uint8_t depth = 0;
     bool in_trust_store = false;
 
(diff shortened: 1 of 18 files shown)


vanetza/security/verify_service.cpp
```diff
@@ -172,6 +172,14 @@ VerifyService straight_verify_service(Runtime& rt, CertificateProvider& cert_pro
         boost::optional<Certificate> signer;
 
         for (auto& cert : possible_certificates) {
+            SubjectType subject_type = cert.subject_info.subject_type;
+
+            if (subject_type != SubjectType::Authorization_Ticket) {
```
riebl: Even a third kind of check besides chain length and time consistency :-)

vanetza/security/naive_certificate_provider.hpp
```diff
@@ -48,6 +48,15 @@ class NaiveCertificateProvider : public CertificateProvider
...
     const Certificate& root_certificate();
 
+    /**
+     * \brief generate a authorization ticket
+     *
+     * \param mutator allows mutating the certificate directly before signing, useful in tests
+     *
+     * \return generated certificate
+     */
+    Certificate generate_authorization_ticket(std::function<void(Certificate&)> mutator);
```
riebl: Hm, this may be a little bit too complicated in the long run. Maybe it is sufficient (for testing) to provide a sign_certificate() method here which enables to forge malicious certificates with correct signature. In my opinion it would not hurt to calculate the signature twice for testing.

vanetza/security/default_certificate_validator.cpp
```diff
@@ -39,46 +29,105 @@ CertificateValidity DefaultCertificateValidator::check_certificate(const Certifi
...
+    m_crypto_backend(backend),
+    m_time_now(time_now),
+    m_trust_store(trust_store),
+    m_cert_cache(cert_cache)
+{
+}
+
+CertificateValidity DefaultCertificateValidator::check_certificate(const Certificate& certificate)
+{
+    return check_certificate(certificate, 10);
```
riebl: Is "10" just a magic number or is there some rationale behind it?

vanetza/security/naive_certificate_provider.hpp
kelunik: Yes, calculating the signature twice totally doesn't hurt, I'll change that.

kelunik:
10 is an arbitrary number, but I think I've read somewhere that chains up to 6 or 8 certificates should be supported. OpenSSL has a similar [default, which is 9](https://github.com/php/php-src/blob/a7fe2570d3ce6915d4ea85c62c0f880ddc225ba7/ext/openssl/php_openssl.h#L41), I had in mind it would be 10, too.

vanetza/security/naive_certificate_provider.hpp
kelunik: Changed.

vanetza/security/default_certificate_validator.cpp
kelunik: It's a magic number, see comment in the regular PR thread.

kelunik:
@riebl I'm not 100% sure about 70bc8d5e1d78a7a7cb9e46049639bdcb945b9d6e, what do you think about that?

riebl:
@kelunik I second avoiding to cast report codes as this could introduce bugs when the enums get changed. In general I would like to keep `DecapRequest` and `DecapConfirm` close to the tables given in TS 102 723-8. Hence, I would also keep `VerifyRequest` and `VerifyConfirm` because they belong to a distinct interface.
Also, your change 70bc8d5 hides the reason why a certificate is considered to be invalid. In my opinion, this reduces the quality of the test assertions.

kelunik:
@riebl Yes, this was also my concern, but `VerifyConfirm` would need to map all the reasons that are anyway present in `DecapConfirm`, as well as copy all data to be available to the facilities layer, which needs those to check SSPs and the ITS-AID.

I think it could make sense to re-add `CertificateInvalidReason`, but drop everything that isn't specific to one certificate being invalid, such as `UNKNOWN_SIGNER` or `TOO_LONG_CHAIN`.

One thing I wasn't sure about is adding `Excessive_Chain` to `DecapReport`, because the interface is kind of standardized by TS 102 723-8, but I'm not even sure why that document specifies these as mandatory etc. instead of only being a recommendation. How it's exactly implemented is an implementation detail after all?

riebl:
I think creating a `DecapConfirm` from a `VerifyConfirm` should be an easy and inexpensive computation by moving the plaintext packet. According to TS 102 723-8, the SN-DECAP.confirm is basically a union of SN-VERIFY.confirm and SN-DECRYPT.confirm: We could model this for example as `using DecapReport = boost::variant<VerificationReport, DecryptionReport>`. I haven't fully thought about this yet though. 

Removing the certificate chain error codes from `CertificateInvalidReason` sounds good to me. Maybe the certificate chain checking should return a custom report data structure?

I don't have objections against adding `Excessive_Chain` to `DecapReport`, it is in fact an implementation detail. I guess it is fine to provide a more powerful interface than the one described in TS 102 723-8.

riebl:
Okay, I have merged parts of this PR now, see fbf5f65. This includes all your changes except 70bc8d5 plus various modifications. For example, the `RectangularRegion` checks were a little bit screwed up.

kelunik:
Was there anything wrong except for the `>` vs `>=` which checks that northwest and southeast aren't equal?

riebl:
Yes, I have "reversed" the comparison operators in some cases, e.g. when comparing the rectangle and position latitudes. Likewise, the two points defined in the unit test were actually "southwest" and "northeast".
Don't worry, though, I don't see my bugs either ;-)

kelunik:
Closing this then, will do a new commit for the non merged one if I have a better idea.

PR #35, opened 2018-01-28
security: Drop security profiles and use ITS-AID instead

kelunik:
This allows setting the correct ITS-AID for GN beacons.


riebl:
Merged 61f830f

PR #40, opened 2018-02-07
security: Fix header serialization order

kelunik:
Sorting is done in the header policy instead of in the serialization function to (1) keep the const in serialize() and (2) keep the order flexible for both testing other implementations and further certificate profiles.

The order is specified to be always ascending, unless overridden by a certificate profile. All currently defined certificate profiles specify that signer_info must be encoded first.


riebl:
Merged 89a772c

PR #43, opened 2018-02-07
security: Fix public key encoding on disk

kelunik:
Previously the EccPoint was encoded only, without the leading byte that specifies the key type. This first byte is necessary to encode a key as it's done in TS 103 097 and by other implementations that expose the public key.


riebl:
Merged 740dd4e

PR #44, opened 2018-02-07
Fix small mistake in README

kelunik:
GeoNetworking is a layer 3 protocol just as IP. Socktap uses GeoNetworking over Ethernet, not over IP. There are no IP addresses involved.


riebl:
Yes, that's all true. However, you can still wrap GeoNet into IP packets though socktap takes another approach. My intention in the README is just to highlight the fact, that Vanetza can be deployed in many different ways.

kelunik:
Well, true. It's basically just tunneling packets, could also happen via UDP.

PR #46, opened 2018-02-10
socktap: Split into CAM sender and dummy BTP sender

kelunik:


riebl:
I would prefer to keep `HelloApplication` as it is and add a new `AwarenessApplication` instead.

kelunik:
I thought about this as well, but I'd like to keep the options handling simple, which gets a lot more complex with additional options for multiple applications. I want to add at least an `--interval` / `--cam-interval` option.

But we might want to just duplicate `main.cpp` for that and maybe skip the security options for the `hello-world.cpp`. That would also allow adding further binaries like a `benchmark.cpp` that sends and signs messages as fast as possible or counts the verified incoming messages per second.

riebl:
So in the end, we both want to keep it simple :-)
I think those application flavours (current socktap, benchmark, hello-world) can share a large portion of socktap's code. Thus, `HelloApplication` could remain but be used only in a KISS hello-world program whereas option handling is kept simple as well for socktap with a single `AwarenessApplication`.

kelunik:
@riebl I've refactored the PR to use two separate binaries instead. I'm open to names, could also be `ca-service` and `hello-world`. We could also get entirely rid of the name `socktap` and rename the directory to `examples` with a sub-directory for all shared code and a separate directory for each demo application.

riebl:
@kelunik For the moment, just leave everything in tools/socktap and add CMake targets for each application flavour (as you have done already).

riebl:
Merged c5f93990875871bf5aa18fcf4b3dbfe15c53ae15.

PR #47, opened 2018-02-14
Fix message signature calculation

kelunik:
This fixes the signature calculation to be compliant with TS 103 097 v1.2.1. This mistake has been detected during interoperability testing with commercial hardware. The captured secured CAM has been added as test case.

Mutual acceptance previously worked, because both sides were doing it equally wrong.


riebl:
Merged 35e3376.

PR #49, opened 2018-02-14
Ensure correct order of certificate elements

kelunik:
They must be encoded in ascending order. Security profiles are allowed to override this order, but no defined security profile does that.


riebl:
Merged 6f918ec

PR #50, opened 2018-02-17
Fix header serialization order

kelunik:
This is a follow-up PR on #40. A unit test is provided first to show that the current code isn't enough to ensure the correct order.


riebl:
Merged with modifications 612dd91

PR #51, opened 2018-02-18
security: Fix chain order

kelunik:
This fixes the certificate order in case a certificate chain is sent or received. See TS 103 097 v1.2.1, section 4.2.10.


riebl:
Merged cdd1886

PR #52, opened 2018-02-18
Check permissions during verification

kelunik:
Permissions are now checked and exposed to the upper layers, so the facilities layer can check the SSPs.

I'm not sure about the `Incompatible_Protocol` reason for messages without any application ID, but there doesn't seem to be a good value defined for that.

`Invalid_Certificate` for certificates without proper permissions for the given message seems fine, but it's a bit weird combined with the fact that `certificate_validity` will be `true`.


riebl:
Merged bfcd89f. I have moved the certificate modification helpers to the `Certificate` itself.

PR #53, opened 2018-02-19
security: Prevent DoS by long certificate chains

kelunik:


riebl:
Merged 686c0c9. Added TODO for custom DoS or "excessive chain length" report code. Since usage of report codes is not specified I see no problem in adding custom ones. Those defined in TS 102 723-8 are just the required minimum set in my view.

kelunik:
Agreed, but before we extend the list, we should model `DecapReport` as `boost::variant<VerificationReport, DecryptionReport>`. I think it also makes sense to add a `Unknown` item as first one, so uninitialized doesn't gives `Success`.

riebl:
I have created #54 for this purpose.

PR #55, opened 2018-02-25
Improve message validation

kelunik:
This PR significantly improves the performance for the validation of incoming messages. As the consistency of certificates is now fully checked (or rejected in places with `TODO` notes), the doesn't need to happen a full chain validation for incoming messages.

A two benchmarks are provided that show a significant speed advantage if messages only include the certificate hash, as only one signature (the message signature) has to be validated in this case instead of the message signature and the certificate signature.

It might be worth to lookup certificates in the cache if we receive a message with a full certificate (same for the chain) for further improvements, but this isn't provided as part of this PR.


riebl:
@kelunik Can you please squash fixup commits? This would ease reviewing a lot.

kelunik:
@riebl The only fixup commit is the last one as far as I can see. Any other commits you'd like to see squashed?

riebl:
If that is the only fixup commit then it's fine. Honestly, I had only a brief look on the log so far.

kelunik:
I've pushed a few fixup commits now, but they only affect the benchmarks. Should I squash them or should I do that later?

riebl:
Merged 5d11621, thanks!

PR #56, opened 2018-03-02
security: Implement is_within for exact matches

kelunik:
Exact matches are trivially within themselves. The implementation of this was surprisingly easy, I hope it's right that way.


riebl:
Please add inequality operators as well (the usual `return !(a == b);` stuff).

kelunik:
@riebl Done. Also fixed the implementation and added a test, as it previously was comparing two `GeographicRegion` instances, which didn't have an overall overloaded implementation.

PR #57, opened 2018-03-05
Drop unsigned packets if security is enabled

kelunik:
Otherwise packets with an invalid timestamp etc. are dropped, but completely unsigned packets aren't. If the facilities layer implements the security checks correctly, I would still notice such packets, but that way it's consistent.


riebl:
Merged c33b7f6

PR #72, opened 2019-02-18
security: implement decompression of ecc points

glmax:
The definition of `EccPoint` in ETSI TS 103 097 V1.2.1, section 4.2.5, allows representing a point on an elliptic curve in a compressed form (only the x coordinate and optional one bit of information about the y coordinate).

This PR adds support for decompression of ECC points for the CryptoPP and OpenSSL crypto backends.


vanetza/security/ecc_point_decompression_visitor.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <vanetza/security/ecc_point.hpp>
+
+namespace vanetza
+{
+namespace security
+{
+
+struct EccPointDecompressionVisitor : public boost::static_visitor<Uncompressed> {
+    Uncompressed operator()(X_Coordinate_Only p) {
+        // Without the additional bit of the y coordinate, we cannot restore y.
```
riebl: Returning incomplete data seems wrong to me.

glmax: The memo you referenced does not use all possible keys/EC points but only those whose y coordinate lies within the lower half of the field, i.e. `y <= (p-1) / 2` resp. `y == min(y, p-y)`. The memo's decompression does not require an additional bit of information in the _compressed form_ of the EC point because the memo defines that information in advance by using only half of the possible EC points.

Since the common representation of (possibly compressed) EC points includes the compression type in the first byte anyway, I don't see a point in omitting the LSB_Y information by using compression type 0 (X_Coordinate_Only).

Concerning our decompression implementation: We cannot make the assumption that all keys we encounter have this property and therefore we cannot decompress *all* keys only from their x coordinate. So either we return incomplete data, throw an error and exit, or throw an error silently (are there any possibilities for that in vanetza?) and the security entity cannot use that key further.

vanetza/security/backend_openssl.cpp
```diff
@@ -86,18 +86,18 @@ class OpenSslEccPointDecompressionVisitor: public EccPointDecompressionVisitor
         const EC_GROUP *group = EC_GROUP_new_by_curve_name(NID_X9_62_prime256v1);
         EC_POINT *point = EC_POINT_new(group);
         BIGNUM *y_coordinate = BN_new();
+
+        Uncompressed p { x };
+        p.y.resize(p.x.size());
 #if OPENSSL_API_COMPAT < 0x10101000L
         EC_POINT_set_compressed_coordinates_GFp(group, point, x_coordinate, static_cast<unsigned>(type) % 2, ctx);
         EC_POINT_get_affine_coordinates_GFp(group, point, nullptr, y_coordinate, ctx);
+        BN_bn2bin(y_coordinate, p.y.data() + (p.y.size() - BN_num_bytes(y_coordinate)));
```
riebl: Please check if `p.y.size()` is actually greater or equal than `BN_num_bytes(y_coordinate)`.
Resize buffer to the larger of both as recovery?

vanetza/security/certificate.cpp
```diff
@@ -102,24 +102,24 @@ void sort(Certificate& cert)
...
-boost::optional<Uncompressed> get_uncompressed_public_key(const Certificate& cert)
+boost::optional<Uncompressed> get_uncompressed_public_key(const Certificate& cert, Backend& backend)
 {
     boost::optional<Uncompressed> public_key_coordinates;
     for (auto& attribute : cert.subject_attributes) {
         if (get_type(attribute) == SubjectAttributeType::Verification_Key) {
             const VerificationKey& verification_key = boost::get<VerificationKey>(attribute);
             const EccPoint& ecc_point = boost::get<ecdsa_nistp256_with_sha256>(verification_key.key).public_key;
-            public_key_coordinates = boost::get<Uncompressed>(ecc_point);
+            public_key_coordinates = backend.decompress_ecc_point(ecc_point);
```
riebl: Catch std::logic_error and return boost::none then.

vanetza/security/ecc_point_decompression_visitor.hpp
riebl: I would rather not call exit or ignore an error state silently. Either throw an exception (if it is really exceptional) or use boost::optional or alike to represent a failure (if it is common that an action fails).

riebl:
@glmax Please have a look at my `decompress_ecc_point` branch. I have refactored your commits and extended the error handling a little bit.

glmax:
Thanks for the refactoring! There's just one small issue: [In this line](https://github.com/riebl/vanetza/blob/f35e795ef3d8c4eec57dda81e82a2a3f6e80863c/vanetza/security/backend_openssl.cpp#L123), the second and third argument for `BN_bn2binpad` need to use `result.y` instead of `p.y`.

Is there any particular reason why you omitted the tests?

riebl:
@glmax Thanks for your feedback, I have fixed the `p.y` bug. Your test suite got accidentally lost during my rebase. I have added it again and refactored it to use a predefined compressed ECC points instead of generating two suitable points at each run.
Thanks a lot for your effort, it is merged now with commit 7e1fe96.

PR #91, opened 2019-09-20
Miscellaneous small improvements for socktap and certify

glmax:
### Certify

Because the GeoNet router automatically sends GN beacons under certain conditions (and if beaconing is not disabled) and because these beacons belong to AID 141 GN-MGMT, an ITS station that wants to sign a beacon needs to have AID 141 and its (empty) SSP vector in its authorization ticket. Hence, this PR adds a command-line option to certify's `generate-ticket` to include the AID/SSP for GN-MGMT.

### Socktap-Common

- Remove duplicate assignment of some attributes.

### Socktap-Cam

- Use the BTP port enum value instead of a magic number.
- Reuse the boost unit cast from geonet to convert lat and lon to the correct representation for CAM generation.

### DIFF
--- a/tools/certify/commands/generate-ticket.cpp
+++ b/tools/certify/commands/generate-ticket.cpp
@@ -31,6 +31,7 @@ bool GenerateTicketCommand::parse(const std::vector<std::string>& opts)
         ("days", po::value<int>(&validity_days)->default_value(7), "Validity in days.")
         ("cam-permissions", po::value<std::string>(&cam_permissions), "CAM permissions as binary string (e.g. '1111111111111100' to grant all SSPs)")
         ("denm-permissions", po::value<std::string>(&denm_permissions), "DENM permissions as binary string (e.g. '000000000000000000000000' to grant no SSPs)")
+        ("include-gn-mgmt", "Include GN-MGMT's SSP (empty)")
     ;
 
     po::positional_options_description pos;
@@ -52,6 +53,8 @@ bool GenerateTicketCommand::parse(const std::vector<std::string>& opts)
 
         return false;
     }
+
+    include_gn_mgmt = vm.count("include-gn-mgmt") > 0;
 
     return true;
 }
@@ -111,6 +114,10 @@ int GenerateTicketCommand::execute()
     certificate_ssp_den.its_aid = IntX(aid::DEN);
     certificate_ssp_den.service_specific_permissions = denm_ssps;
     certificate_ssp.push_back(certificate_ssp_den);
+
+    if (include_gn_mgmt) {
+        certificate_ssp.push_back({IntX(aid::GN_MGMT), vanetza::ByteBuffer{}});
+    }
 
     certificate.signer_info = calculate_hash(sign_cert);
     certificate.subject_info.subject_type = SubjectType::Authorization_Ticket;
--- a/tools/certify/commands/generate-ticket.cpp
+++ b/tools/certify/commands/generate-ticket.cpp
@@ -31,7 +31,7 @@ bool GenerateTicketCommand::parse(const std::vector<std::string>& opts)
         ("days", po::value<int>(&validity_days)->default_value(7), "Validity in days.")
         ("cam-permissions", po::value<std::string>(&cam_permissions), "CAM permissions as binary string (e.g. '1111111111111100' to grant all SSPs)")
         ("denm-permissions", po::value<std::string>(&denm_permissions), "DENM permissions as binary string (e.g. '000000000000000000000000' to grant no SSPs)")
-        ("include-gn-mgmt", "Include GN-MGMT's SSP (empty)")
+        ("permit-gn-mgmt", "Generated ticket can be used to sign GN-MGMT messages (e.g. beacons).")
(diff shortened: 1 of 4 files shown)


tools/certify/commands/generate-ticket.cpp
```diff
@@ -31,6 +31,7 @@ bool GenerateTicketCommand::parse(const std::vector<std::string>& opts)
         ("days", po::value<int>(&validity_days)->default_value(7), "Validity in days.")
         ("cam-permissions", po::value<std::string>(&cam_permissions), "CAM permissions as binary string (e.g. '1111111111111100' to grant all SSPs)")
         ("denm-permissions", po::value<std::string>(&denm_permissions), "DENM permissions as binary string (e.g. '000000000000000000000000' to grant no SSPs)")
+        ("include-gn-mgmt", "Include GN-MGMT's SSP (empty)")
```
kelunik: I guess it should be `gn-mgmt-permission` for consistency.

glmax: I just thought `gn-mgmt-permission(s)` to be misleading because you cannot specify any permissions, but I don't mind.

kelunik: That's also true, maybe just `--gn-mgnt`?

glmax: `--with-gn-mgmt` ?

glmax: > `--with-gn-mgmt` ?

@kelunik, are you also okay with that?

kelunik: @glmax, don't really care. I'll let @riebl decide.

tools/socktap/cam_application.cpp
```diff
@@ -76,8 +67,8 @@ void CamApplication::on_timer(const boost::system::error_code& ec)
     BasicContainer_t& basic = cam.camParameters.basicContainer;
     basic.stationType = StationType_passengerCar;
     basic.referencePosition.altitude.altitudeValue = AltitudeValue_unavailable;
-    basic.referencePosition.longitude = round(position.longitude, microdegree) * Longitude_oneMicrodegreeEast;
-    basic.referencePosition.latitude = round(position.latitude, microdegree) * Latitude_oneMicrodegreeNorth;
+    basic.referencePosition.longitude = std::round(static_cast<geo_angle_i32t>(position.longitude).value());
+    basic.referencePosition.latitude = std::round(static_cast<geo_angle_i32t>(position.latitude).value());
```
riebl: Why do you prefer this approach?
I do not see the advantage at the moment and the previous code lines look more readable to me.

tools/socktap/gps_position_provider.cpp
```diff
@@ -76,9 +76,6 @@ void GpsPositionProvider::fetch_position_fix()
         fetched_position_fix.longitude = gps_data.fix.longitude * degree;
         fetched_position_fix.speed.assign(gps_data.fix.speed * si::meter_per_second, gps_data.fix.eps * si::meter_per_second);
         fetched_position_fix.course.assign(north + gps_data.fix.track * degree, north + gps_data.fix.epd * degree);
-        fetched_position_fix.timestamp = convert(gps_data.fix.time);
-        fetched_position_fix.latitude = gps_data.fix.latitude * degree;
-        fetched_position_fix.longitude = gps_data.fix.longitude * degree;
```
riebl: Good catch!

tools/certify/commands/generate-ticket.cpp
riebl: I have no strong preference either, however, `--allow-gn-mgmt` would make it clear that it is related to permissions.

tools/socktap/cam_application.cpp
glmax: I thought it would be better to re-use the existing conversion of the degree representation. However, I just realized that after casting to a `int32_t` quantity, the round should be a noop or am I missing something? Is it readability-wise okay when the round is omitted?

Apart from that, it's probably even better to move all code to extract/convert information from the `PositionFix` into a `CAM` to `cam_permissions.hpp/cpp`.

tools/certify/commands/generate-ticket.cpp
riebl: I am totally convinced now :-)

tools/socktap/cam_application.cpp
riebl: Yes, calling `std::round` after casting to an integer type has no effect.
I am not a fan of calling `value()` on those `boost::units::quantity<>` because the result depends on the stored unit's scale. Just imagine what happens if GeoNetworking decides to use a different scale than the CDD… happy debugging then!
That's the reason why `microdegree` is passed in the present code: Making things explicit.

A helper function `to_reference_position(const PositionFix&): ReferencePosition_t` would be less verbose, agreed.

riebl:
Merged except for "socktap-cam: use boost unit cast for longitude and latitude"

glmax:
The other changes have been moved to #98.

PR #92, opened 2019-09-20
socktap: update to Cohda LLC network interface

glmax:
Cohda has discontinued the WAVE network interfaces on the MK5 devices some time ago and added 'Cohda LLC' network interfaces. When using raw sockets like before, these interfaces expect Cohda LLC interface messages with 802.11p frames that are almost ready for transmission (sequence number and frame control sequence (CRC32) is inserted/added by the interface).

This PR replaces the old MK2Tx/RxDescriptors in socktap by the corresponding Cohda LLC interface messages and adds generation and parsing of the packet contents below GN, namely the LLC header (**logical link control, not the same as the Cohda LLC interface/API!**) and the 802.11 header.

The Cohda LLC interface/API is available for Cohda Release 15 firmware and later. Furthermore, the old WAVE implementation in socktap might no longer work for Release 15 firmware and later.

### DIFF
--- a/tools/socktap/cohda.cpp
+++ b/tools/socktap/cohda.cpp
@@ -1,41 +1,149 @@
 #include "cohda.hpp"
-#include <vanetza/dcc/data_request.hpp>
-#include <cassert>
-#include <mk2mac-api-types.h>
+#include <boost/crc.hpp>
+#include <iostream>
+#include <llc-api.h>
+#include <vanetza/geonet/router.hpp>
+#include <vanetza/geonet/address.hpp>
+#include <vanetza/dcc/profile.hpp>
+#include <vanetza/common/byte_order.hpp>
 
-namespace vanetza
-{
+namespace vanetza {
 
-using dcc::Profile;
+struct LLCHeader {
+    uint8_t DSAP = 0xAA;
+    uint8_t SSAP = 0xAA;
+    uint8_t ControlField = 0x03;
+    uint8_t OUI[3] = { 0x00, 0x00, 0x00 };
+    uint16_t ProtocolID = geonet::ether_type.net();
+};
 
-ByteBuffer create_cohda_tx_header(const dcc::DataRequest& request)
-{
-    ByteBuffer header(sizeof(tMK2TxDescriptor), 0x00);
-    assert(header.size() == sizeof(tMK2TxDescriptor));
-    tMK2TxDescriptor* tx = reinterpret_cast<tMK2TxDescriptor*>(header.data());
-    tx->ChannelNumber = 180;
+/* See ETSI EN 302 663 V1.2.1 (2013-07), Table B.3 */
+enum tPriority {
+    AC_BK = 1, AC_BE = 3, AC_VI = 5, AC_VO = 6
+};
+constexpr uint8_t QOS_PRIO_MASK = 0x07;
+
+/**
+ * 1 bit payload type (MSDU = 0)
+ * 2 bits ack policy (No_Ack = 1)
+ * 1 bit EOSP (service period = 0)
+ * 1 bit unused
+ * 3 bits prioriy (set dynamically according to traffic class)
+ * 8 bits TXOP (shall be set to 0 when communicating outside of a BSS)
+ */
+struct tQOSControl {
+    uint8_t QOSFlags = 0x20;
+    uint8_t TXOP = 0; // not used
+};
+
+struct tIEEE80211PHeader {
+    uint8_t FrameControlVersionAndType = 0x88;
+    uint8_t FrameControlFlags = 0x00; // not used
+    uint16_t DurationID = 0x00; // not used
+    uint8_t Destination[6];
+    uint8_t Source[6];
+    uint8_t BSSID[6] = { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff }; // Operation outside of a BSS, (see ETSI EN 302 663 V1.2.1 Annex A)
+    uint16_t SequenceControl = 0x00; // 12 bits for sequence number, 4 bits for fragment number; will be set by Cohda LLC
+    tQOSControl QOSControl;
+};
+
+struct tTXPhysical {
+    tMKxTxPacket TxMessage;
+    tIEEE80211PHeader IEEE80211PHeader;
+};
+
+constexpr uint8_t IEEE80211P_FCS_LENGTH = 4;
+constexpr uint32_t CRC32_MAGIC_NUMBER = 0x2144df1c;
+
+struct tRXPhysical {
+    tMKxRxPacket RxMessage;
+    tIEEE80211PHeader IEEE80211PHeader;
+};
+
+void insert_cohda_tx_header(const dcc::DataRequest& request, std::unique_ptr<ChunkPacket>& packet) {
+    LLCHeader llcData;
+    auto llcPtr = reinterpret_cast<uint8_t*>(&llcData);
+    packet->layer(OsiLayer::Link) = std::move(ByteBuffer(llcPtr, llcPtr + sizeof(LLCHeader)));
+
+    tTXPhysical phyData;
+    phyData.TxMessage = {};
+    std::copy(request.destination.octets.begin(), request.destination.octets.end(), phyData.IEEE80211PHeader.Destination);
+    std::copy(request.source.octets.begin(), request.source.octets.end(), phyData.IEEE80211PHeader.Source);
     switch (request.dcc_profile) {
-        case Profile::DP0:
-            tx->Priority = 7; // AC_VO
+        case dcc::Profile::DP0:
+            phyData.IEEE80211PHeader.QOSControl.QOSFlags |= QOS_PRIO_MASK & tPriority::AC_VO; // AC_VO (Voice)
             break;
-        case Profile::DP1:
-            tx->Priority = 5; // AC_VI
+        case dcc::Profile::DP1:
+            phyData.IEEE80211PHeader.QOSControl.QOSFlags |= QOS_PRIO_MASK & tPriority::AC_VI; // AC_VI (Video)
             break;
-        case Profile::DP2:
-            tx->Priority = 3; // AC_BE
+        case dcc::Profile::DP2:
+            phyData.IEEE80211PHeader.QOSControl.QOSFlags |= QOS_PRIO_MASK & tPriority::AC_BE; // AC_BE (Best effort)
             break;
-        case Profile::DP3:
-            tx->Priority = 1; // AC_BK
+        case dcc::Profile::DP3:
+            phyData.IEEE80211PHeader.QOSControl.QOSFlags |= QOS_PRIO_MASK & tPriority::AC_BK; // AC_BK (Background)
             break;
         default:
-            tx->Priority = MK2_PRIO_NON_QOS;
+            phyData.IEEE80211PHeader.QOSControl.QOSFlags |= QOS_PRIO_MASK & tPriority::AC_BE;
             break;
     }
-    tx->Service = MK2_QOS_ACK;
-    tx->MCS = MK2MCS_DEFAULT;
-    tx->TxPower.PowerSetting = MK2TPC_DEFAULT;
-    tx->TxAntenna = MK2_TXANT_DEFAULT;
-    return header;
+
+    uint16_t payload_size = sizeof(tIEEE80211PHeader) + packet->size(OsiLayer::Link, OsiLayer::Application);
+    uint16_t total_size = sizeof(tMKxTxPacket) + payload_size;
+
+    phyData.TxMessage.Hdr.Type = MKXIF_TXPACKET;
+    phyData.TxMessage.Hdr.Len = total_size;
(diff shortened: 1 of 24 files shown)


riebl:
Looks good to me. I would like to make some tests on Mk5 before merging it, however.

tools/socktap/cohda.cpp
```diff
@@ -1,41 +1,149 @@
...
 
-ByteBuffer create_cohda_tx_header(const dcc::DataRequest& request)
-{
-    ByteBuffer header(sizeof(tMK2TxDescriptor), 0x00);
-    assert(header.size() == sizeof(tMK2TxDescriptor));
-    tMK2TxDescriptor* tx = reinterpret_cast<tMK2TxDescriptor*>(header.data());
-    tx->ChannelNumber = 180;
+/* See ETSI EN 302 663 V1.2.1 (2013-07), Table B.3 */
+enum tPriority {
+    AC_BK = 1, AC_BE = 3, AC_VI = 5, AC_VO = 6
```
riebl: This table is a good hint. However, AC_BE is 0 then.
I am also thinking about changing `net::AccessCategory`...

tools/socktap/cohda.cpp
```diff
@@ -1,41 +1,149 @@
...
+struct tIEEE80211PHeader {
+    uint8_t FrameControlVersionAndType = 0x88;
+    uint8_t FrameControlFlags = 0x00; // not used
+    uint16_t DurationID = 0x00; // not used
+    uint8_t Destination[6];
+    uint8_t Source[6];
+    uint8_t BSSID[6] = { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff }; // Operation outside of a BSS, (see ETSI EN 302 663 V1.2.1 Annex A)
+    uint16_t SequenceControl = 0x00; // 12 bits for sequence number, 4 bits for fragment number; will be set by Cohda LLC
+    tQOSControl QOSControl;
+};
```
riebl: We might want to add this to the `net` directory where also our `EthernetHeader` is located.

tools/socktap/cohda.cpp
```diff
@@ -1,41 +1,149 @@
...
+namespace vanetza {
 
-using dcc::Profile;
+struct LLCHeader {
+    uint8_t DSAP = 0xAA;
+    uint8_t SSAP = 0xAA;
+    uint8_t ControlField = 0x03;
+    uint8_t OUI[3] = { 0x00, 0x00, 0x00 };
+    uint16_t ProtocolID = geonet::ether_type.net();
+};
```
riebl: Candidate for a `llc_snap_header.hpp`

tools/socktap/cohda.cpp
```diff
@@ -1,41 +1,149 @@
...
+
+    boost::crc_32_type crc_result;
+    /*
+     * Process the whole 802.11p frame (including its FCS). If the FCS is correct, the CRC result has the same value
+     * as the CRC of one block (32 bit) of zeroes.
+     */
+    crc_result.process_bytes(packet.buffer().data() + sizeof(tMKxRxPacket), packet.size() - sizeof(tMKxRxPacket));
+    if (crc_result.checksum() != CRC32_MAGIC_NUMBER) {
+        return boost::none;
+    }
```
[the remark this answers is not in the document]
glmax: Oh, great, thanks!

tools/socktap/cohda.cpp
glmax: Since 802.11 allows values 0 and 3 for AC_BE, I would prefer 3, because it's the larger value and larger than AC_BK values, so a total order is maintained. Consequently, AC_VO should be changed to 7 for the same reasons that it's the larger of both possible values.

riebl: Hm, maintaining a total order is a good point. I thought it would be nice then to keep a 1:1 relation of the names with 802.1d. Anyway, the numbering is odd and should not bother us too much in the end.

glmax: So which one do you prefer:
- 1, 0, 5, 6 with a comment referencing the table and pointing out compatability to 802.1D, or
- 1, 3, 5, 7 with a commend referencing the table and pointing out the BE/BK resolution in 802.11?

riebl: 1, 3, 5, 7 looks good

glmax:
When moving the LLC and 802.11p header into separate headers files, I would still like to combine them into one link-layer structure because the ETSI standards describe the ITS-G5 link layer as consisting of LLC+802.11p (MAC header). With that in mind, wouldn't they rather belong into something like `vanetza/access/itsg5_link_layer.hpp`?

riebl:
Yes, `vanetza/access` makes more sense. Unfortunately, `net` and `access` lack some consistency. For example, `AccessCategory` and `EthernetHeader` should be in `access` as well then.
I am struggling with the name `itsg5_link_layer.hpp` because it looks quite ugly, doesn't it? `its_link_layer.hpp` might be to broadly defined, however?

glmax:
`its_link_layer` is indeed to broadly defined. What about `g5_link_layer.hpp`?

riebl:
I am fine with `g5_link_layer.hpp`, it won't get any better I guess.

glmax:
@riebl, do you wish any further changes?

glmax:
Rebased the branch on the current master and resolved the conflicts in `flow_control.hpp`/`tests/flow_control.cpp`

riebl:
I am currently experimenting with your branch in my Cohda VM. No big issues so far :+1: 
Hopefully, I can run a quick test on two MK5 in our lab this afternoon. I will then merge your branch. Sorry for all the delays!

riebl:
Okay, I have added some changes on top of your branch. I would like to ask you @glmax to review them. You can find them in my `socktap_cohda_llc` branch.
I have tested this code with Cohda MK5 and Autotalks Craton2 units and it looks good.

In a nutshell, I would like to avoid casting byte buffers to structs so I have added serialization code for `access::G5LinkLayer`.
As far as I know, ACK policy is not required to be "No Ack" but "Normal Ack" is fine as well.

glmax:
All in all, it looks good (and pretty, with the (de)serialization ;)). 👍 

I'd like to discuss only some minor points.

---------------

> As far as I know, ACK policy is not required to be "No Ack" but "Normal Ack" is fine as well.

My reasoning behind the ACK policy was that we ignore the packet because we don't actually send the acknowledgement, but it's probably okay to accept the packet anyway.

---------------

About the [frame control fields](https://github.com/riebl/vanetza/commit/72e7707831f4b8992e21910f0a965c39675f4c1d#diff-53159c9883260054416a74c1cbb46851R34):

- More Fragments: When we accepted fragmented frames, wouldn't we be required to puzzle incoming fragments together? As far as I know, at least the Cohda MK5 does not use fragmentation on TX (although I don't know whether it can receive fragmented frames). I will ask the support for details and give an update here.
- Retry: If this flag is set but we also received the previous TX attempt(s) successfully, will the GN router filter out this duplicate(/s)?

---------------

About the arm compiler:

I know you don't like gcc-4.8, but at least at this point, one can get gcc-4.8 to compile vanetza simply by using round instead of curly brackets in [this initialization](https://github.com/riebl/vanetza/blob/master/vanetza/security/region.cpp#L681).

With gcc-5.4, I get linker errors which I haven't been able to resolve up to now:

- When building shared libs, I get this error when `cohda_proxy` and `proxy_fake_feed` are linked:  
    ```
    ../../lib/libvanetza_common.so: undefined reference to `boost::gregorian::greg_month::get_month_map_ptr[abi:cxx11]()'
    ```
- When building static libs, I get tons of linker errors for Crypto++ and Boost.Program_Options when `benchmark` is linked.

riebl:
My motivation of being less strict regarding ACK policy is simply to be nice to third-party senders which may enable ACKs, e.g. for greedy forwarding.

I hope we don't have to worry much about fragmentation and retransmissions, but that Cohda's MAC layer handels this for us. In deed, duplicates are (partially) filtered by GeoNetworking. Do you think it could hurt to be less strict regarding the presence of those flags?

Just yesterday I thought about removing `cohda_proxy` and `proxy_fake_feed` entirely. Is anybody using these ancient tools at all?
Those errors are strange nevertheless: I guess you are also using Cohda's VM for this? Why do we see different outcomes with (almost) identical build systems?

glmax:
> Do you think it could hurt to be less strict regarding the presence of those flags?

*If* Cohda LLC handles fragmentation, we *shouldn't* see any signs of fragmentation in the Frame Control Field or the Sequence Control Field (that would make the most sense, I think). If Cohda LLC doesn't handle fragmentation we either need to implement it (at least for RX) or be strict with `More Fragments` and probably also `Fragment Number`. However, I will give feedback once I've heard back from the support.

-------

> Just yesterday I thought about removing `cohda_proxy` and `proxy_fake_feed` entirely. Is anybody using these ancient tools at all?

When refactoring the link layer code into `vanetza/access`, I actually wondered whether it is necessary to adapt `cohda_proxy` to Cohda LLC, but I think your idea to remove both tools answers this question quite good. ;)

-------

> Those errors are strange nevertheless: I guess you are also using Cohda's VM for this? Why do we see different outcomes with (almost) identical build systems?

I just tested it with a completely fresh Cohda SDK Rel16 and did the following (the results were exactly the same as described above):

- clone vanetza
- download and unpack vanetza arm dependencies from box.com
- build and install CMake 3.14.5
- `sudo apt-get autoremove gcc-4.8-arm-linux-gnueabihf`
- edit `${VANETZA}/cmake/Toolchain-Cohda-MK5.cmake`:  
    ```
    set(CMAKE_C_COMPILER "arm-linux-gnueabihf-gcc-5")
    set(CMAKE_CXX_COMPILER "arm-linux-gnueabihf-g++-5")
    ```
    (the package with gcc-5.4 does not provide symlinks for gcc and g++ without the major version)
- invoke cmake with the following options:  
    ```
    cmake ${VANETZA} \
              -DCMAKE_TOOLCHAIN_FILE=${VANETZA}/cmake/Toolchain-Cohda-MK5.cmake \
              -DCMAKE_FIND_ROOT_PATH=${VANETZA_MK5_DEPS} \
              -DCMAKE_INSTALL_RPATH=\$ORIGIN/../lib \
              -DGPS_LIBRARY=/home/duser/mk5/stack/v2x-lib/lib/mk5/libgps_static.a \
              -DGPS_INCLUDE_DIR=/home/duser/mk5/stack/v2x-lib/include \
              -DCMAKE_EXE_LINKER_FLAGS="-pthread" \
              -DBUILD_SOCKTAP=TRUE \
              -DBUILD_BENCHMARK=TRUE \
              -DBUILD_SHARED_LIBS=TRUE \
              -DBUILD_CERTIFY=TRUE
    ```
- invoke make (I `make` with `` `nproc` `` jobs, but that doesn't affect the linker problems, it's just nice to be much faster)

glmax:
Concerning the fragmentation: Cohda does not support it because it's not required by 802.11p. Therefore, I'd rather add `More Fragments` and `Fragment Number` to the list of "fixed fields".

riebl:
Agreed, I have updated my branch accordingly so all Frame Control bits are fixed now.

I have revised the build instructions for Cohda SDK, e.g. `CMAKE_EXE_LINKER_FLAGS` seems no longer to be necessary. Also, I have uploaded a new set of pre-built dependencies and those are working fine with *benchmark*, *socktap* and *certify* enabled for shared and static builds.
Usually, I am just too lazy to copy all shared objects to MK5 and thus prefer static builds…

The `arm-linux-gnueabihf-g++` symlinks are provided by the `g++-arm-linux-gnueabihf` package, as far as I know.

glmax:
> Also, I have uploaded a new set of pre-built dependencies and those are working fine with _benchmark_, _socktap_ and _certify_ enabled for shared and static builds.

I can confirm that. 👍 

> The `arm-linux-gnueabihf-g++` symlinks are provided by the `g++-arm-linux-gnueabihf` package, as far as I know.

Indeed, you're right. My "problem" was that by default (i.e. like shipped by Cohda), the 4.8 versions of the tools are symlinked and when (auto)removing `g++-4.8-arm-linux-gnueabihf` the links are deleted. In order to have the symlinks to version 5 of the tools (without doing it manually), one can just autoremove `g++-{,4.8-}arm-linux-gnueabihf` and then install `g++-arm-linux-gnueabihf` again.

>  Usually, I am just too lazy to copy all shared objects to MK5 and thus prefer static builds…

Since the built libraries are installed along with the binaries during `make install` anyway, I like the smaller total size (although it doesn't make too much of a difference with `-O3`).

> Agreed, I have updated my branch accordingly so all Frame Control bits are fixed now.

Final remark: The `Fragment Number` should also be required to be zero. Otherwise we would accept the last fragment of a fragmented packet.

riebl:
I have just updated the *vanetza-deps* archive and added a `libcryptopp.so.8` symlink. For now, I don't consider the wrongly reported Boost version as a show-stopper.
Also, the fragment number is checked and the documentation fixed, thanks for your contribution and valuable feedback! Finally merged :-)

PR #95, opened 2019-11-08
Add transmit/drop hooks in flow_control reporting destination ports

kenog:
For some of our simulations we needed to distinguish packets dropped by the gatekeeper. The existing drop/transmit hooks are only useful if all running applications have different access categories which does not necessarily have to be the case. This downward compatible PR adds hooks reporting the BTP-B destination port which should indeed always be different for different applications. It also wraps the calls of the hooks, keeping the code of the gate keeper's core funtions tidy -- even if more hooks would be added in the future.

### DIFF
--- a/vanetza/dcc/CMakeLists.txt
+++ b/vanetza/dcc/CMakeLists.txt
@@ -17,7 +17,7 @@ set(CXX_SOURCES
 )
 
 add_vanetza_component(dcc ${CXX_SOURCES})
-target_link_libraries(dcc PUBLIC access btp net)
+target_link_libraries(dcc PUBLIC access net)
 
 add_test_subdirectory(tests)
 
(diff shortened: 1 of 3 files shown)


riebl:
I agree that this is a handy feature. However, I see two problems:

1. `dcc::FlowControl` should be generic but this change binds it to BTP-B
2. `parse_btp_b` is a non-const operation but `FlowControl` should not modify any packet

I think that passing a const packet reference to a `Hook` would be fine, though. It is then the callback's responsibility to determine the particular attribute (such as BTP-B port) from the packet.

kenog:
I fully agree - good idea! I hope I adapted the code according to what you meant. Do you think we should keep the `call_{transmit,dop}_hooks `functions? Right now they do not really make a difference. I added them merely not to clutter the actual functionality of the `FlowControl` entity. I head in mind that further functions, e.g. reporting other BTP/GN/IP... fields , could be added in the future. However, with your suggested changes, this is probably not relevant anymore.

riebl:
I have merged your idea with 59918f5. If you have any interesting "packet dissecting functions" that would be helpful for others to write hook callbacks, I would be happy to merge them as well!
Thanks a lot!

PR #98, opened 2019-11-26
socktap: create helper function to copy position information into CAM

glmax:
This PR moves the code that fills a CAM's `ReferencePosition` with the information from a `PositionFix` to `cam_functions.hpp/.cpp` and reuses `cam_functions`'s unit for latitude/longitude in CAMs. Also, see [this comment](https://github.com/riebl/vanetza/pull/91#discussion_r330111263).


riebl:
Merged, thanks!

PR #103, opened 2020-02-05
decode_application

erikbritto:
The added decode_application.cpp gets the broadcasted CAM and displays each CAM content via COUT. 

The socktap-decode executable is generated together with all socktap applications.

### DIFF
--- a/tools/socktap/CMakeLists.txt
+++ b/tools/socktap/CMakeLists.txt
@@ -1,4 +1,4 @@
-if(NOT TARGET Boost::system)
+f(NOT TARGET Boost::system)
     message(STATUS "Skip build of socktap because of missing Boost::system dependency")
     return()
 endif()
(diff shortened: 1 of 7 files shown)


riebl:
Thanks for your pull request @erikbritto! I like your extension to print the content of received CAMs.
Nevertheless, I have a few remarks :-)
- printing a CAM should be moved to a function so it can be reused, e.g. a `void print(std::ostream&, const asn1::Cam&)` added to cam_functions.hpp
- I would like to avoid yet another socktap executable but add this printing feature to `main_cam.cpp`. It should be easy to enable printing in `CamApplication` then, e.g. with a new `void CamApplication::enableOutput(bool)` controlled by a command line option.
- May it improve readability of the printed CAM when tabulator indentation is used instead of prefixing every field with its "field path"?

riebl:
I have merged a modified version of your code, see the new `print_intended` function in *cam_facilities.hpp*. Thanks for your PR!

PR #105, opened 2020-02-24
Additions to cross-layer interfaces according to the ETSI standards

glmax:
This PR adds several fields/"parameters" to the primitives that are passed between

- facilities/application and btp (BTP-DATA.request/indication, [ETSI ITS 302 636-5-1, Annex A](https://www.etsi.org/deliver/etsi_en/302600_302699/3026360501/02.02.01_60/en_3026360501v020201p.pdf))
- btp and geonet (GN-DATA.request/indication, [ETSI ITS 302 636-4-1, Annex J](https://www.etsi.org/deliver/etsi_en/302600_302699/3026360401/01.03.01_60/en_3026360401v010301p.pdf))
- geonet and the security entity (SN-ENCAP.request/confirm, SN-DECAP.request/confirm, [ETSI ITS 102 723-8, sections 5.2.12 and 5.2.13](https://www.etsi.org/deliver/etsi_ts/102700_102799/10272308/01.01.01_60/ts_10272308v010101p.pdf))

It also changes the datatype of the source position in BTP/GN-DATA.indication.

Together with the introduction of the permissions field for the request primitives comes the possibility to check whether the permissions requested by the facilities/application (derived from the payload) match the available permissions in the authorization ticket used for signing the message.


vanetza/geonet/router.cpp
```diff
@@ -229,8 +235,14 @@ DataConfirm Router::request(const ShbDataRequest& request, DownPacketPtr payload
...
 
             // step 2: encapsulate packet by security
-            if (m_mib.itsGnSecurity) {
-                payload = encap_packet(ctrl.its_aid, *pdu, std::move(payload));
+            if (m_mib.itsGnSecurity && ctrl.security_profile) {
+                auto encap_result = encap_packet(ctrl.security_profile, ctrl.its_aid, *ctrl.permissions, *pdu, std::move(payload));
+                if (encap_result) {
+                    payload = std::move(encap_result.get());
+                } else {
+                    // TODO: How to throw an error here?
```
riebl: As far as I know, no error reporting is foreseen at this point by specifications.
We may invoke a hook here so custom error logging can be realised without hassle.

vanetza/geonet/router.cpp
```diff
@@ -1209,26 +1236,32 @@ std::unique_ptr<GbcPdu> Router::create_gbc_pdu(const GbcDataRequest& request)
     return pdu;
 }
 
-Router::DownPacketPtr Router::encap_packet(ItsAid its_aid, Pdu& pdu, DownPacketPtr packet)
+boost::optional<Router::DownPacketPtr> Router::encap_packet(boost::optional<security::SecurityProfile> security_profile, ItsAid its_aid, ByteBuffer permissions, Pdu& pdu, DownPacketPtr packet)
```
riebl: return an empty pointer instead of a boost::optional?

vanetza/security/encap_request.hpp
```diff
@@ -9,13 +9,29 @@ namespace vanetza
...
+    // TODO: Add Encrypt and SignAndEncrypt when encryption is supported
+};
+
 /**
  * Input data for encapsulating a packet in a secured message
  * \see TS 102 723-8 for SN-ENCAP.request
  */
 struct EncapRequest {
     DownPacket plaintext_payload; // mandatory
+    boost::optional<SecurityProfile> sec_services; // optional
```
riebl: I know that *sec_services* is defined in TS 102 723-8 but it does not make any sense if we are not going to implement encryption. Do you have plans to implement encryption? Is anybody using this feature at all?
Since `SecurityProfile::Default` exists, I see no point in wrapping it into a boost::optional.

vanetza/security/permission_check.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace {
+/**
+ * Most SSP definitions consist of two types of bits:
+ * - Bits that must match exactly in request and AT, e.g. the version number part, identifiers (like with IVIM) or reserved bits
+ * - Bits that grant individual permissions and can only be successfully requested if they are set in the AT
+ *
+ * For easy bytewise SSP checking, one can use bitmasks to mark where which type of check is required.
+ *
+ * This map contains such a bitmask for all supported applications per supported SSP version of the respective application.
```
riebl: Can you add some references please?

vanetza/security/tests/security_entity.cpp
```diff
@@ -720,20 +739,21 @@ TEST_F(SecurityEntityTest, verify_message_without_position_and_without_restricti
...
+//TEST_F(SecurityEntityTest, verify_message_with_insufficient_aid)
+//{
+//    its_aid = 42; // some random value not present in the certificate
+//
+//    // verify message
+//    DecapConfirm decap_confirm = security.decapsulate_packet(DecapRequest { create_secured_message() });
+//    EXPECT_EQ(DecapReport::Invalid_Certificate, decap_confirm.report);
+//    ASSERT_FALSE(decap_confirm.certificate_validity);
+//    EXPECT_EQ(CertificateInvalidReason::Insufficient_ITS_AID, decap_confirm.certificate_validity.reason());
+//}
```
riebl: Please fix this test case.

vanetza/btp/data_indication.hpp
```diff
@@ -34,6 +35,7 @@ struct DataIndication
     geonet::LongPositionVector source_position;
     geonet::TrafficClass traffic_class;
     boost::optional<geonet::Lifetime> remaining_packet_lifetime;
+    boost::optional<unsigned> remaining_hop_limit;
```
riebl: I suppose `remaining_packet_lifetime` and `remaining_hop_limit` are optional to implement but always available from a GN protocol perspective. What do you think about making these two fields "mandatory" in Vanetza then?

PR #106, opened 2020-02-24
Miscellaneous small improvements for socktap, btp and geonet

glmax:
This PR is a collection of different small improvements for btp, geonet, socktap.

- BTP: Added hooks can now be removed again.
- GN: The security report is only initialized when the security entity has actually been invoked.
- Socktap:
    - Applications can be disabled again (combined with the BTP change, one can close ports in both directions).
    - Status messages are assembled separately and then printed to `std::cout` as one string.
    - AA certificates provided via commandline are put into the certificate cache so if other stations use the same AA, the certificate does not need to be requested from them.
    - New command line parameter to control MIB parameter `ItnGnSnDecapResultHandling`.
    - Make the `SecurityEntity` parameter of `RouterContext`'s constructor a pointer to conform with the interface/requirements of `Router`.

### DIFF
--- a/tools/socktap/dcc_passthrough.cpp
+++ b/tools/socktap/dcc_passthrough.cpp
@@ -6,6 +6,7 @@
 #include <vanetza/net/chunk_packet.hpp>
 #include <boost/asio/generic/raw_protocol.hpp>
 #include <iostream>
+#include <sstream>
 
 #ifdef SOCKTAP_WITH_COHDA_LLC
 #include "cohda.hpp"
@@ -43,7 +44,9 @@ void DccPassthrough::request(const dcc::DataRequest& request, std::unique_ptr<Ch
         const_buffers[i] = asio::buffer(buffers_[i]);
     }
     auto bytes_sent = socket_.send(const_buffers);
-    std::cout << "sent packet to " << request.destination << " (" << bytes_sent << " bytes)\n";
+    std::ostringstream ostream;
+    ostream << "sent packet to " << request.destination << " (" << bytes_sent << " bytes)\n";
+    std::cout << ostream.str();
 }
 
 void DccPassthrough::allow_packet_flow(bool allow)
--- a/tools/socktap/main_cam.cpp
+++ b/tools/socktap/main_cam.cpp
@@ -43,6 +43,7 @@ int main(int argc, const char** argv)
         ("print-rx-cam", "Print received CAMs")
         ("print-tx-cam", "Print generated CAMs")
         ("benchmark", "Enable benchmarking")
+        ("non-strict", "Set MIB parameter ItsGnSnDecapResultHandling to NON_STRICT")
     ;
 
     po::positional_options_description positional_options;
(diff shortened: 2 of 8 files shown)


tools/socktap/dcc_passthrough.cpp
```diff
@@ -43,7 +44,9 @@ void DccPassthrough::request(const dcc::DataRequest& request, std::unique_ptr<Ch
         const_buffers[i] = asio::buffer(buffers_[i]);
     }
     auto bytes_sent = socket_.send(const_buffers);
-    std::cout << "sent packet to " << request.destination << " (" << bytes_sent << " bytes)\n";
+    std::ostringstream ostream;
+    ostream << "sent packet to " << request.destination << " (" << bytes_sent << " bytes)\n";
+    std::cout << ostream.str();
```
riebl: Why is this beneficial? From my point of view, this causes an additional memory allocation and thus this should be justifiable.

vanetza/geonet/router.cpp
```diff
@@ -398,7 +398,6 @@ void Router::indicate_basic(IndicationContextBasic& ctx)
             indicate_secured(ctx, *basic);
         } else if (basic->next_header == NextHeaderBasic::Common) {
             if (!m_mib.itsGnSecurity || SecurityDecapHandling::Non_Strict == m_mib.itsGnSnDecapResultHandling) {
-                indication.security_report = security::DecapReport::Unsigned_Message,
```
riebl: No, `security_report` is not optional in `geonet::DataIndication` though EN 302 636-4-1 says so. Why do you think that `Unsigned_Message` is unsuitable?

glmax: Oh, sorry, that commit/change needs to be moved to #105. There, `security_report` is made optional because of EN 302 636-4-1 and EN 302 636-5-1.

tools/socktap/dcc_passthrough.cpp
glmax: When Vanetza shares stdout with other threads that manipulate the flags for stdout, it can happen (although rarely and unpredictable) that the `std::hex` flag set in the streaming operator overload of `MacAddress` is not properly cleared. One could probably move the special treatment to the streaming operator overload.

If you don't see that as a problem Socktap should care about, feel free to omit this commit.

riebl: Okay, but socktap is not multi-threaded ;-) If this is going to change, we will need some more sophisticated logging anyway.

tools/socktap/main_cam.cpp
```diff
@@ -144,6 +148,7 @@ int main(int argc, const char** argv)
                 for (auto& chain_path : vm["certificate-chain"].as<std::vector<std::string> >()) {
                     auto chain_certificate = security::load_certificate_from_file(chain_path);
                     chain.push_back(chain_certificate);
+                    cert_cache.insert(chain_certificate);
```
glmax: Regarding the comment below: Demonstrating certificate requests by using tickets signed by different AAs is a cleaner and closer to the use of this feature in the real world.

If you want to keep this need for a request for the own AA certificate as a demo feature, I would propose that it is off by default and can be activated via a command line option.

tools/socktap/main_cam.cpp
```diff
@@ -144,6 +148,7 @@ int main(int argc, const char** argv)
...
 
                     // Only add root certificates to trust store, so certificate requests are visible for demo purposes.
```
glmax: Regarding the change above: Demonstrating certificate requests by using tickets signed by different AAs is a cleaner way and closer to the use of this feature in the real world.

If you want to keep this need for a request for the own AA certificate as a demo feature, I would propose that it is off by default and can be activated via a command line option.

riebl:
Merged except for "status message as single string" and the not-yet optional `security_report`.

PR #107, opened 2020-02-25
Fix and new feature (dual alpha) for LIMERIC 

QuentinDel:
Hello :) 
This pull request consists on two different things:
- The fix: Avoid to keep updating m_interval when this one is already "expired". The problem comes when the variable delay is lower than 0, the m_interval is not consistent anymore.

- The new feature: To reduce the convergence time of LIMERIC in transition period, a paper (https://ieeexplore.ieee.org/document/8671471) proposes to use a Dual-Alpha approach. This is what I implemented.

Regards,
Q.


riebl:
@QuentinDel and I have discussed his `m_interval` fix and I am still not convinced that it is the right thing to do. A more thorough investigation is needed.

riebl:
Dual-alpha feature has been merged now.

riebl:
@QuentinDel has been absolutely right and I am convinced now after spending some time on this issue :-)
Thanks and merged.

PR #113, opened 2020-11-20
Return left hand side in move assignment operator for `asn1c_wrapper_common`

hagau:
The move assignment operator of `asn1c_wrapper_common` doesn't return a value, but should return the left hand side.


riebl:
Merged addc2fe5fc213b6d63a1ce0a8749eb935216afa3. Thanks @hagau for fixing this stupid bug!

PR #114, opened 2020-11-26
Update BTP port numbers to latest version of the standard

kenog:
Hi Raphael,

in ETSI TS 103 248 V1.3.1 (2019-04) there are a couple of new BTP ports defined. This PR adds them to vanetza.

Note: The names of the according variables resembled the names of the messages rather than the names of the services before. I stuck to that pattern to keep it consistent. The downside is that there are now five different variables for port 2011 - one for each message of the TPG service.
The other option would be to name the ports after the services but that breaks backwards compatibility and is not nice because the TLC has two messages with different ports for the two different messages.
I found the current solution the better trade-off but am happy to change it if you prefer the second option.

Best regards
Keno

### DIFF
--- a/vanetza/btp/ports.hpp
+++ b/vanetza/btp/ports.hpp
@@ -13,11 +13,28 @@ typedef uint16be_t port_type;
 namespace ports
 {
 
+// Port numbers according to ETSI TS 103 248 v1.3.1 (2019-04)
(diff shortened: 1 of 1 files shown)


riebl:
Hi Keno,

I am fine with your trade-off. Adding more aliases for ports should be no issue anyway.
Thanks for your pull request.

Regards,
Raphael

PR #119, opened 2020-12-18
cmake: fix build with GeographicLib source code

jlesk:
Hello,
I would like to build Vanetza with GeographicLib source code.
In my project Vanetza and GeographicLib are both submodules.
I manage to do it with this patch but I am not sure this is the correct way.
Please, if my method is not correct, could you give a sample CMakeLists.txt file to build
both projects?
Thank you
Best regards.
PS: I am using ubuntu 18.04 and geographicLib 1.49.2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.1)
 project(Vanetza VERSION 0.8)
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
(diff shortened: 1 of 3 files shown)


riebl:
Hi @jlesk,

AFAIK, the usage of library variables (such as *${GeographicLib_LIBRARIES}*) is discouraged in modern CMake in favour of imported targets (such as *GeographicLib::GeographicLib*).
I also somewhat doubt that the modification of the *CMAKE_MODULE_PATH* variable is necessary because we can incorporate Vanetza without problems in Artery via `add_subdirectory` as it is.
Right now, I am not convinced to merge your pull request. If you can bring up good reasons, however, I may change my mind.

riebl:
Stale.

PR #129, opened 2021-05-14
Update to Cohda SDK R17 and upper version

bma67:
Just a small fix to compile with Cohda SDK R17.x


riebl:
Does this break compatibility with older Cohda SDK releases?

bma67:
Yes, it does because Cohda simply renamed a variable member from a structure.

Sanjaer:
I encountered this problem too, there is a change in the names of several fields in mk2mac-api-types.h

riebl:
Obsolete by now because I have removed the tools *proxy_fake_feed* and *cohda_proxy*. They have never been closely related to Vanetza after all.

PR #132, opened 2021-07-08
Update router.cpp

oscarmex1986:
This is a bug that affects forwarded packets, since DCC cannot read a value for lifetime and defaults to zero, effectively killing packets in the DCC queues.


riebl:
Thanks a lot @oscarmex1986! I have merged your fix as ec54488023c009636250e8d4d22744a94a5768a1 with a reworded commit message.

Issue #139, opened 2022-01-11
Adding new message types


buraksenkus:
Hi,

I want to be able to use new message types (e.g. SPAT, MAP). I have the ASN files of these messages but I don't know how to proceed. Should I add a new class for each field in the messages under the asn1 folder? If so should I do this manually? Can you help me with this?

Best regards,
Burak

riebl:
Hi @buraksenkus,

You can add your ASN.1 files to the asn1 directory and include them to the CMake rules located in https://github.com/riebl/vanetza/blob/master/vanetza/asn1/CMakeLists.txt. If you have asn1c installed on your computer, a "generate_asn1c" target is enabled by Vanetza that will generate C code out of the ASN.1 files.
At the moment, I am using a slightly patched version of asn1c which you can find here: https://github.com/riebl/asn1c/tree/integer_uper
The classes in the vanetza/asn1 folder are just for convenience and not strictly required to encode or decode ASN.1 data structures.

Do you need released versions of SPAT and MAP? If yes, it would be great to add them to Vanetza's main branch.

buraksenkus:
Hi again,

I've tried as you said but .c files are not created. Here is my steps:

1. I've installed asn1c from https://github.com/riebl/asn1c/tree/integer_uper repo.
2. I've added my ASN.1 files into asn1 folder.
![image](https://user-images.githubusercontent.com/19664856/149491521-30bd9348-0a7f-48c5-be55-a6390cf690d4.png)

2. I've included my files in `vanetza/asn1/CMakeLists.txt` and added some messages for looking whether it is worked or not.
![image](https://user-images.githubusercontent.com/19664856/149491864-6744ae5b-988e-491b-9b8e-013e1181bb33.png)

3.  I've run cmake with `-DVANETZA_ASN1_WITH_ASN1C=ON ..` flag and everything has gone correctly.
![image](https://user-images.githubusercontent.com/19664856/149492375-c1673768-b699-4fde-91a6-c579e3bd0530.png)

4. Finally, I've run `make` and `make install` but no new files are created inside `vanetza/asn1/its` folder.

I couldn't understand what I'm missing or doing wrong.

riebl:
Invoking `make generate_asn1c` should do the trick. Since the Vanetza repository ships with pre-generated C sources of the ASN.1 stuff for convenience, the build system does not automatically regenerate those files every time. You need to trigger this step explicitly.

Otherwise, you seem to be on the right track! :-)

[...]

PR #140, opened 2022-01-19
SPATEM & MAPEM support

buraksenkus:
Regarding the issue I've opened (#139),

- ASN files have been added.
   - ASN files were taken from [here](http://oldforge.etsi.org/websvn/listing.php?repname=ITS.ITS_ASN1).
   - There were some duplicates in `ISO_TS_19091_CPM.asn` and `DSRC_REG_D.asn` files so one of them had to be changed. Therefore, the `ISO_TS_19091_CPM.asn` file was removed and missing fields such as **VehicleHeight** were added into the `DSRC_REG_D.asn`.
- Generation operation has been performed and classes are created.
- Related wrapper header files have been added.

SPATEM and MAPEM messages are ready to use.


riebl:
Dear @buraksenkus,

Thanks a lot for sharing your efforts! Are you sure those ASN.1 definitions reflect the latest message revision? I think https://forge.etsi.org/rep/ITS/asn1/is_ts103301 has more up-to-date versions of SPATEM and MAPEM.

buraksenkus:
Dear @riebl,

I think you're right. I unintentionally didn't check the repository you mentioned. I'll update PR when I'm available.

buraksenkus:
Hi @riebl,

I've updated the PR using the latest ASN.1 definitions from the repository you specified. I think there is no problem other than that. I'd be happy to help if there's anything else that needs fixing.

Note: Since the Heading, HeadingConfidence and SpeedConfidence definitions in the `ISO-TS-19091-addgrp-C-2018.asn` file on [this](https://standards.iso.org/iso/ts/19091/ed-2/en/) site conflict with the definitions in the ITS-Container module, I added **DSRC** suffix to them since it is done in this way in the given repository.

buraksenkus:
I just noticed an **Reg_NodeOffsetPointXY_t** error during the build operation. I will try to solve it.

buraksenkus:
Finally I've solved the problem and updated the PR. Now, ASN.1 definitions are the latest ones.

In this comment https://github.com/riebl/vanetza/pull/140#issuecomment-1033680563, I said that I updated the ASN.1 files to the latest version, but I also encountered the circular dependency problem mentioned [here](https://github.com/vlm/asn1c/issues/272). I don't know if the problem is caused by defining ASN.1 files or by asn1c, but with the help of @pitzor, I added the **-fno-include-deps** flag to asn1c command and the problem has been solved. This caused changes in many files although most of them were not functional.

Also, I had to change the cam_functions.hpp/cpp a little bit because of include statements.

I've tested the content of the MAPEMs in Wireshark and it works as expected.

image

I wouldn't want to change the existing codes too much, but I couldn't solve the situation without doing them. I hope there is nothing that I have missed.

riebl:
Thanks for your efforts @buraksenkus! I will have a closer look as soon as possible.

riebl:
@buraksenkus Can you please review my *mapem_spatem* branch? Does it fulfil your needs?

buraksenkus:
It fulfils my needs. Thanks for your effort @riebl!

riebl:
Thanks for your feedback @buraksenkus. I have merged the MAPEM and SPATEM messages into the *master* branch.

PR #150, opened 2022-04-12
socktap: fetch altitude for GPSD_API_MAJOR_VERSION > 8

berkayyaman:
"altitude" variable is deprecated after GPSD_API_MAJOR_VERSION = 8.
Rather "altHAE" (height above ellipsoid) or "altMSL" (mean sea level) is used.

According to ETSI TS 102 894-2 V1.3.1 (2018-08), height above ellipsoid is used in data frames.


riebl:
Thanks for your pull request! Based on your commit, I have added a `gpsd_get_altitude` function, so these `#ifdef` blocks dealing with various GPSd versions are separated from the actual logic.

PR #151, opened 2022-04-13
socktap: Adding the Autotalks link layer

khevessy:
This adds Autotalks link layer to the socktap example, working on both Craton and Secton devices. Links with Autotalks libraries. Also adds possibility to build for Craton (cross-compilation for ARM using the toolchain supplied with Autotalks SDK).

Initialization of the device can be taken from the Autotalks example, as I propose in the documentation page (I used almost exactly their code, therefore I think it cannot be in the library).

I should add that my knowledge of CMake is not great, so that part may require some further work.


tools/socktap/link_layer.cpp
```diff
@@ -30,6 +35,12 @@ create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& dev
         ip::udp::endpoint multicast(ip::address::from_string("239.118.122.97"), 8947);
         link_layer.reset(new UdpLink { io_service, multicast });
     }
+    else if (name == "autotalks") {
+#ifdef SOCKTAP_WITH_AUTOTALKS
+        v2x_device_init();
```
riebl, line 40: move `v2x_device_init` to `AutotalksLink` constructor

tools/socktap/main.cpp
```diff
@@ -86,10 +90,14 @@ int main(int argc, const char** argv)
...
         }
 
-        auto signal_handler = [&io_service](const boost::system::error_code& ec, int signal_number) {
+        auto signal_handler = [&io_service, &link_layer](const boost::system::error_code& ec, int signal_number) {
             if (!ec) {
                 std::cout << "Termination requested." << std::endl;
                 io_service.stop();
+#ifdef SOCKTAP_WITH_AUTOTALKS
+                if (nullptr != dynamic_cast<AutotalksLink*>(link_layer.get()))      // Deinitialize Autotalks device only if its link layer used
+                    autotalks_device_deinit();
```
riebl, line 99: Move `autotalks_device_deinit` to the (yet missing) destructor of `AutotalksLink`. Then no changes to *main.cpp* are necessary at all.

tools/socktap/autotalks.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <vanetza/net/cohesive_packet.hpp>
+#include "stddef.h"
+#include "atlk/sdk.h"
+#include "atlk/v2x_service.h"
+#include "autotalks_link.hpp"
+
+/*
+ * Device initialization.
+ */
+int v2x_device_init(void);
```
riebl, line 18: This function is called `v2x_device_init` but the corresponding deinitialization function is prefixed "autotalks". This is strange.

tools/socktap/autotalks.cpp
```diff
@@ -0,0 +1,229 @@
...
+  #ifdef SECTON
+    rc = ref_sys_init_ex(2, (char**) arg);
+  #else
+    #error Wrong configuration selected.
+  #endif
+#endif
+
+  // TODO                                                                     //
+  // TODO: Take initialization from Autotalks basic example's main function   //
+  // TODO                                                                     //
```
riebl, line 56: Proposal: Add `#error "Autotalks device initialization code is missing"` so this code does not silently compile and dysfunctional application.

tools/socktap/autotalks.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * Parse information from received Autotalks packet.
+ */
+boost::optional<vanetza::EthernetHeader> strip_autotalks_rx_header(vanetza::CohesivePacket&, v2x_receive_params_t);
+
+/*
+ * Create a new thread for receiving data.
+ */
+void init_rx(AutotalksLink*);
+
+
```
riebl, line 60: Please wrap these functions in a namespace, e.g. `vanetza::autotalks`.

cmake/Toolchain-Autotalks.cmake
```diff
@@ -0,0 +1,15 @@
...
+set(CMAKE_C_FLAGS "--sysroot=/opt/poky-craton2/2.6.2/sysroots//cortexa7t2hf-neon-poky-linux-gnueabi -march=armv7ve -mthumb -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a7" CACHE STRING "" FORCE)
+set(CMAKE_CXX_FLAGS "--sysroot=/opt/poky-craton2/2.6.2/sysroots//cortexa7t2hf-neon-poky-linux-gnueabi -I/opt/poky-craton2/2.6.2/sysroots//cortexa7t2hf-neon-poky-linux-gnueabi/usr/include  -march=armv7ve -mthumb -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a7" CACHE STRING "" FORCE)
+set(THREADS_PTHREAD_ARG "2" CACHE STRING "Forced by Autotalks toolchain" FORCE)
+# Cache string needed in cmake v3.5.1, in v3.16.3 not anymore
+# THREADS_PTHREAD_ARG needed for a bug that is also not present in v3.16.3 anymore
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+set(AUTOTALKS_CRATON true)
```
riebl, line 14: So this toolchain is actually *Toolchain-Autotalks-Craton.cmake*?

cmake/Toolchain-Autotalks.cmake
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_SYSTEM_NAME "Linux")
+set(CMAKE_SYSTEM_PROCESSOR "arm")
+set(CMAKE_C_COMPILER "/opt/poky-craton2/2.6.2/sysroots/x86_64-pokysdk-linux/usr/bin/arm-poky-linux-gnueabi/arm-poky-linux-gnueabi-gcc")
```
riebl, line 3: Is */opt/poky-craton2* a path you have chosen or is it common for all Autotalks SDK installations?
Anyhow, I suggest adding a POKY_CRATON2 CMake variable with "/opt/poky-craton2/2.6.2" so one can easily switch to another toolchain (version).

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,91 @@
...
+
+
+    foreach(dir ${AUTOTALKS_INCLUDE_DIRS})
+        message("Directory: " ${dir})
+    endforeach()
```
riebl, line 86: Is this loop a leftover from debugging?

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,91 @@
...
..._linux_u.o ${AUTOTALKS_OBJ_DIR}/link_layer/linux_u/local/link_layer_local.o ${AUTOTALKS_OBJ_DIR}/link_layer/linux_u/ipc/link_layer_ipc.o ${AUTOTALKS_OBJ_DIR}/link_layer/linux_u/tee/static_list.o ${AUTOTALKS_OBJ_DIR}/link_layer/linux_u/tee/link_layer_tee.o ${AUTOTALKS_OBJ_DIR}/link_layer/link_layer_interface.o ${AUTOTALKS_OBJ_DIR}/flash_reader/linux_u/local/flash_reader_local.o ${AUTOTALKS_OBJ_DIR}/ref_sys/ref_sys.o ${AUTOTALKS_OBJ_DIR}/ref_sys/ref_sys_logger.o ${AUTOTALKS_OBJ_DIR}/log_reader/log_reader.o ${AUTOTALKS_OBJ_DIR}/secure_hdif_crypto_layer/crypto_layer_stubs.o ${AUTOTALKS_OBJ_DIR}/time_sync/time_sync.o ${AUTOTALKS_OBJ_DIR}/time_sync/time_sync_poti.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/gnss.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_io.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_io_check.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_nav.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_parse.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/teseo/teseo.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/teseo/teseo_commands.o ${AUTOTALKS_OBJ_DIR}/poti_lib/coord/coord.o ${AUTOTALKS_OBJ_DIR}/poti_lib/periodic_alarm/periodic_alarm.o ${AUTOTALKS_OBJ_DIR}/poti_lib/api/poti_api.o)
```
riebl, line 81: These are not *executable* but *objects*

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,91 @@
...
+
+else() # Craton device
+    find_path(
+        AUTOTALKS_ROOT
+        NAMES "include/atlk/v2x_service.h"
+        PATHS $ENV{HOME}
```
riebl, line 42: So you expect the API to be located at */home/your_user/autotalks_craton_api*? Is this suggested by Autotalks?

tools/socktap/CMakeLists.txt
```diff
@@ -27,7 +27,23 @@ add_executable(socktap
...
 )
 
-target_link_libraries(socktap PUBLIC Boost::system Boost::program_options Threads::Threads vanetza)
+option(SOCKTAP_WITH_AUTOTALKS "Use Autotalks API for socktap" OFF) # Both Secton and Craton devices
+if (SOCKTAP_WITH_AUTOTALKS)
+    find_package(Autotalks MODULE REQUIRED)
+    target_compile_definitions(socktap PUBLIC "SOCKTAP_WITH_AUTOTALKS")
+    if(NOT AUTOTALKS_CRATON)
+        target_compile_definitions(socktap PUBLIC "SECTON")
+        target_link_libraries(socktap PUBLIC Boost::system Boost::program_options Threads::Threads vanetza Autotalks::AtlkRemote Autotalks::ToMcrypt Autotalks::ToMmath ${AUTOTALKS_EXECUTABLES} rt)
```
riebl, line 36: Looks like *Autotalks::AtlkRemote* depends on *ToMcrypt* and *ToMath*? You could specify this dependency in *FindAutotalks.cmake* and also add the *AUTOTALKS_EXECUTABLES* (objects) to the *AtlkRemote* target.
Ideally, you just state *Autotalks::AtlkRemote* as dependency here and everything else (including rt) is then transitively added by CMake.

tools/socktap/autotalks_link.cpp
```diff
@@ -0,0 +1,47 @@
+#include "autotalks_link.hpp"
+#include "autotalks.hpp"
+#include <vanetza/net/osi_layer.hpp>
+
+const unsigned SendBufferSize = 1000;
```
riebl, line 5: The MTU is greater than 1000 bytes. I suggest allocating 2000 bytes on the stack to be on the safe side.

riebl (changes requested):
Thanks for your effort! Unfortunately, I cannot test your code on an actual Autotalks device as I have no longer access to any.
I have added several comments where I would like to see changes before merging it into the master branch. Please don't be shocked by these comments, I really appreciate your contribution!

tools/socktap/autotalks.hpp
Irisbus: Renamed to `autotalks_device_init()`

cmake/Toolchain-Autotalks.cmake
Irisbus: Yes, I renamed it.

cmake/Toolchain-Autotalks.cmake
Irisbus: This is actually a path that is created when installing the toolchain (Autotalks have they own install script), so it should be common for all installations. I have added the POKY_CRATON2 variable nonetheless.

cmake/FindAutotalks.cmake
Irisbus: Removed.

cmake/FindAutotalks.cmake
Irisbus: Renamed.

cmake/FindAutotalks.cmake
Irisbus: I expect the user to create symbolic link from _/home/your_user/autotalks_craton_api_ or _/home/your_user/autotalks_secton_api_ to SDK root, I thought that it was simpler as there does not seem to be any consistent path defined by autotalks. However, the SDK itself is found in `secton-sdk-5.14.0-rel_sec_USB` or `secton-sdk-5.14.0-rel_sec_USB` for version `5.14.0`, so maybe that could be used somehow?

tools/socktap/CMakeLists.txt
Irisbus: Turns out _AtlkRemote_ does not even need _ToMcrypt_ and _ToMmath_. However, I moved the dependencies to _FindAutotalks.cmake_ as you suggested, I thinks now it is correct.

Irisbus:
I am not sure if you are happy with user manually creating the symbolic link to SDK root as mentioned above, but to me it seems good enough (as the path is not defined exactly by Autotalks).
Apart from that I hopefully solved all your requests.

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,78 @@
...
+    find_path(
+        AUTOTALKS_ROOT
+        NAMES "include/atlk/v2x_service.h"
+        PATH_SUFFIXES "autotalks_secton_api"
+        PATHS $ENV{HOME}
+        CMAKE_FIND_ROOT_PATH_BOTH # Seems like this is not needed in newer CMake
+    )
+
+    set(AUTOTALKS_INCLUDE_DIRS "${AUTOTALKS_ROOT}/include" "${AUTOTALKS_ROOT}/ref_src/include" "${AUTOTALKS_ROOT}/depend/device/include" "${AUTOTALKS_ROOT}/src/include" "${AUTOTALKS_ROOT}/src/core/include")
+    set(AUTOTALKS_LIBS_DIR ${AUTOTALKS_ROOT}/output.sec/x86/lib)
```
riebl, line 11: Are there only x86 libraries? May we want to check `CMAKE_SYSTEM_PROCESSOR` to select the right directory?

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,78 @@
...
+        AUTOTALKS_ROOT
+        NAMES "include/atlk/v2x_service.h"
+        PATHS $ENV{HOME}
+        PATH_SUFFIXES "autotalks_craton_api"
+        CMAKE_FIND_ROOT_PATH_BOTH # Seems like this is not needed in newer CMake
+    )
+
+    set(AUTOTALKS_INCLUDE_DIRS "${AUTOTALKS_ROOT}/include" "${AUTOTALKS_ROOT}/ref_src/include" "${AUTOTALKS_ROOT}/depend/device/include" "${AUTOTALKS_ROOT}/src/include" "${AUTOTALKS_ROOT}/src/core/include")
+
+    set(AUTOTALKS_LIBS_DIR ${AUTOTALKS_ROOT}/output.cr2_lnx/armv7-32/lib)
```
riebl, line 39: Here it is always *arm7-32* because that is simply the Craton processor's architecture, isn't it?
Edit: see also *Toolchain-Autotalks-Craton.cmake* which pins the architecture to ARM.

cmake/FindAutotalks.cmake
```diff
@@ -0,0 +1,78 @@
...
...TOTALKS_OBJ_DIR}/link_layer/linux_u/tee/static_list.o ${AUTOTALKS_OBJ_DIR}/link_layer/linux_u/tee/link_layer_tee.o ${AUTOTALKS_OBJ_DIR}/link_layer/link_layer_interface.o ${AUTOTALKS_OBJ_DIR}/flash_reader/linux_u/local/flash_reader_local.o ${AUTOTALKS_OBJ_DIR}/ref_sys/ref_sys.o ${AUTOTALKS_OBJ_DIR}/ref_sys/ref_sys_logger.o ${AUTOTALKS_OBJ_DIR}/log_reader/log_reader.o ${AUTOTALKS_OBJ_DIR}/secure_hdif_crypto_layer/crypto_layer_stubs.o ${AUTOTALKS_OBJ_DIR}/time_sync/time_sync.o ${AUTOTALKS_OBJ_DIR}/time_sync/time_sync_poti.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/gnss.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_io.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_io_check.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_nav.o ${AUTOTALKS_OBJ_DIR}/poti_lib/nmea/nmea_parse.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/teseo/teseo.o ${AUTOTALKS_OBJ_DIR}/poti_lib/gnss/teseo/teseo_commands.o ${AUTOTALKS_OBJ_DIR}/poti_lib/coord/coord.o ${AUTOTALKS_OBJ_DIR}/poti_lib/periodic_alarm/periodic_alarm.o ${AUTOTALKS_OBJ_DIR}/poti_lib/api/poti_api.o ${AUTOTALKS_OBJ_DIR}/poti_lib/ubx/serial_driver.o ${AUTOTALKS_OBJ_DIR}/poti_lib/ubx/ubx_driver.o ${AUTOTALKS_OBJ_DIR}/poti_lib/ubx/ubx_nav.o)
```
riebl, line 63: You can improve readability by using `list(TRANSFORM AUTOTALKS_OBJECTS PREPEND ${AUTOTALKS_OBJ_DIR}/)` and omit the repeated prefix in the `set` call.

tools/socktap/autotalks.cpp
```diff
@@ -0,0 +1,251 @@
...
+bool endRxThread = false;
+
+static pthread_t v2x_rx_thread;
+static uint8_t v2x_rx_buffer[2048];
+
+
+int autotalks_device_init(void)
+{
+  atlk_rc_t rc;
+  const char* arg[] = {"./test", "enx0002ccf00006"};
```
riebl, line 47: black magic?

cmake/FindAutotalks.cmake
riebl: Would you please add a note about this expectation to *autotalks-sdk-build.md*?

riebl (changes requested):
I think I will merge this shortly :-)

tools/socktap/autotalks.cpp
khevessy: I have put the "enx0002ccf00006" network name in define and added a note to the documentation. I think it is hard-coded somewhere from Autotalks

cmake/FindAutotalks.cmake
khevessy: By `x86`, Autotalks code generation means that the system's native compiler is used (e.g. no cross-compilation) so in this case this will really be x86 on every platform (last year I tried SDK compilation on Raspberry PI and libraries were still in x86 folder). Autotalks documentation mentions that when Secton is controlled from Craton and we are cross-compiling for Craton, output folder is armv7-32. However, I did not try this possibility.

cmake/FindAutotalks.cmake
khevessy: Yes, I think this will always stay the same as you say.

cmake/FindAutotalks.cmake
khevessy: It was already there but I tried to make the explanation more clear

khevessy:
Note that `socktap` now does not work as stated [here](https://github.com/riebl/vanetza/issues/192#issuecomment-1671666667). After recent security changes, it cannot be compiled in the base repository either.

riebl:
Sorry, I did not notice that *socktap* cannot be compiled. I should add *socktap* to the CI!
Are you also suffering from the assertion failure mentioned by @lo-hei?

riebl:
Thanks for your hard work and patience @khevessy!

khevessy:
After your fix, I've tried `socktap` on Secton and Craton and it is working well.

PR #156, opened 2022-04-26
Update how-to-build.md

Irisbus:
I thought that your script for updating CMake version could be mentioned somewhere, ignore this if you think it is not needed.


riebl:
Thanks for your valuable suggestion to update the docs accordingly! I prefer to redirect the readers to Kitware's APT repository, though. I have updated the docs accordingly based on your suggestion.

PR #158, opened 2022-04-27
Update ports.hpp and its_aid.hpp

Irisbus:
Update BTP port numbers to latest version of the standard - in ETSI TS 103 248 v2.1.1 there are two more ports defined.
Also, in ETSI TS 102 965 v2.1.1, there is one new ITS-AID for IMZ Service.


riebl:
Thanks @Irisbus, I have merged your PR with b78665fe.

PR #170, opened 2022-10-14
Adds alternative name for lib to FindGeographiclib

kenog:
Hi @riebl,
I installed GeographicLib from the Arch User Repository on my system. Unfortunately, it is seems to be named differently now, such that `cmake/FindGeographicLib.cmake` cannot find it anymore.
This small patch adds the name the library has on my system and CMake now finds the library again.
I thought this might help others, too.

Best regards
Keno


riebl:
Thanks @kenog !

PR #173, opened 2022-11-09
Cam printing enhancements

Irisbus:
Enhanced function `print_indented()` in vanetza/facilities/cam_functions.cpp. Now it should be printing everything if I did not forget something.


Irisbus:
I think that `NULL` would be more correct for null pointer checks instead of `nullptr` because `vanetza::asn1::allocate()` uses `calloc()`, but I am not sure it matters.

riebl:
Thanks, I have added your changes as 3157b5100554f7cc7985cb46d4a6af174c4a9a52.

PR #176, opened 2022-12-01
conan: same dependency versions in conanfile and cmakelists

hsynydn:
This PR includes a simple change in conanfile.py. **_requires_** replaced with **_requirements_** method alternative which could provide a better logic if necessary. The versions of the dependencies changed in a way that give the same meaning with CMakeLists.txt find_package versions.  I have also added into the conanfile.py that the build options which are already exist in CMakeLists.txt.


riebl:
Great, thanks!
On my system I still have minor issue, though: The conan build still picks up the non-conan GeographicLib.

PR #161, opened 2022-05-06
Adding SREM, SSEM and IVIM messages

Irisbus:
I've added SREM and SSEM messages. I think I have it correct, only thing I am not sure with are the files EXTERNAL.c and EXTERNAL.h, my asn1c compiler keeps generating those but I think they may not be needed. Even after removing `/usr/local/share/asn1c`, when I reinstall the asn1c from `riebl/asn1c/integer_uper`, they appear again.

PR #180, opened 2022-12-28
srem_ssem_ivim branch: Rename IssuerIdentifier_t to IssuerIdentifierIso_t (collision with security module). Remove rename of Temperature_t as asn1c now does it automatically.

Irisbus:
Changes as discussed in #161: IssuerIdentifier_t already defined in IEEE 1609.2 which is used by security, I have had some runtime issues with that.
Furthermore, `Temperature_t` renaming is now not needed as used version of asn1c distinguishes `ITS_Container_Temperature_t` and `IVI_Temperature_t`.


riebl:
Merged by bf7a58875d7317f22fc880690ddd12c364b80415

PR #189, opened 2023-05-11
common: Clock epoch_debug() , timestamp in micros

mikelngelous:
In case there are multiple receptions from different services, it is a good idea to know when they arrive, even at the different layers, here is an example in the RouterContext

### DIFF
--- a/tools/pki/README.md
+++ b/tools/pki/README.md
@@ -43,9 +43,9 @@ Prints the DC URL as found in the ECTL for the previously selected Root CA.
 
 `pki dc getctl --print`
 
-> - EA: http://0.eu-ea.l0.c-its-pki.eu/ [AA]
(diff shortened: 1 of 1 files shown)


riebl:
I like your idea @mikelngelous, I am just not convinced that the string formatting should become a method of `vanetza::Clock`. Also, calling `std::chrono::high_resolution_clock` could be quite confusing in the end because "clock ticking" is solely handled by `vanetza::Runtime` everywhere else.

riebl:
Sorry for my late reply; I have been away for a few days. I prefer to have pull requests per feature, i.e. your UDP feature for socktap would be a separate pull request.
What is the reason that you have closed your PR @mikelngelous?

mikelngelous:
It is reopened again.

PR #191, opened 2023-06-07
Added TCP LinkLayer to socktap

fysch:
I've added TCP to socktap.
To accept incoming connections, use `--tcp-accept 127.0.0.1:12345`.
To connect to another service, use `--tcp-connect 127.0.0.1:12345`.

### DIFF
--- a/tools/socktap/link_layer.cpp
+++ b/tools/socktap/link_layer.cpp
@@ -1,5 +1,6 @@
 #include "link_layer.hpp"
 #include "raw_socket_link.hpp"
+#include "tcp_link.hpp"
 #include "udp_link.hpp"
 #include <vanetza/access/ethertype.hpp>
 #include <boost/asio/generic/raw_protocol.hpp>
@@ -9,7 +10,7 @@
 #endif
 
 std::unique_ptr<LinkLayer>
-create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& device, const std::string& name)
+create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& device, const std::string& name, const boost::program_options::variables_map& vm)
 {
     std::unique_ptr<LinkLayer> link_layer;
 
@@ -29,7 +30,43 @@ create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& dev
         namespace ip = boost::asio::ip;
         ip::udp::endpoint multicast(ip::address::from_string("239.118.122.97"), 8947);
         link_layer.reset(new UdpLink { io_service, multicast });
+    } else if (name.substr(0, 3) == "tcp") {
+        namespace ip = boost::asio::ip;
+
+        TcpLink* tcp = new TcpLink { io_service };
+
+        std::string tcp_ip;
+        unsigned short tcp_port, tcp_ip_len;
+
+        if (vm.count("tcp-connect")) {
+            for (const std::string& ip_port : vm["tcp-connect"].as<std::vector<std::string>>()) {
+                tcp_ip_len = ip_port.find(":");
+                tcp_ip = ip_port.substr(0, tcp_ip_len);
+                tcp_port = std::stoi(ip_port.substr(tcp_ip_len + 1));                
+                tcp->connect({ip::address::from_string(tcp_ip), tcp_port});
+            }
+        }
+
+        if (vm.count("tcp-accept")) {
+            for (const std::string& ip_port : vm["tcp-accept"].as<std::vector<std::string>>()) {
+                tcp_ip_len = ip_port.find(":");
+                tcp_ip = ip_port.substr(0, tcp_ip_len);
+                tcp_port = std::stoi(ip_port.substr(tcp_ip_len + 1));
+                tcp->accept({ip::address::from_string(tcp_ip), tcp_port});
+            }
+        }
+
+        link_layer.reset(tcp);
+
     }
 
     return link_layer;
+}
+
+void add_link_layer_options(boost::program_options::options_description& options)
+{
+    options.add_options()
+        ("tcp-connect", boost::program_options::value<std::vector<std::string>>()->multitoken(), "Connect to TCP-Host(s). Comma separated list of [ip]:[port].")
+        ("tcp-accept", boost::program_options::value<std::vector<std::string>>()->multitoken(), "Accept TCP-Connections. Comma separated list of [ip]:[port].")
+    ;
 }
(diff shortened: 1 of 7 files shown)


tools/socktap/link_layer.cpp
```diff
@@ -29,7 +30,43 @@ create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& dev
         namespace ip = boost::asio::ip;
         ip::udp::endpoint multicast(ip::address::from_string("239.118.122.97"), 8947);
         link_layer.reset(new UdpLink { io_service, multicast });
+    } else if (name.substr(0, 3) == "tcp") {
```
riebl, line 33: Why not match the full `name`?

tools/socktap/link_layer.cpp
```diff
@@ -29,7 +30,43 @@ create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& dev
...
+                tcp_ip_len = ip_port.find(":");
+                tcp_ip = ip_port.substr(0, tcp_ip_len);
+                tcp_port = std::stoi(ip_port.substr(tcp_ip_len + 1));                
+                tcp->connect({ip::address::from_string(tcp_ip), tcp_port});
+            }
+        }
+
+        if (vm.count("tcp-accept")) {
+            for (const std::string& ip_port : vm["tcp-accept"].as<std::vector<std::string>>()) {
+                tcp_ip_len = ip_port.find(":");
```
riebl, line 52: I would like to extract the string parsing of IP address and port to a separate function. In this function, we could also do proper error handling. For example, is there a default port if the colon ":" is missing?

tools/socktap/link_layer.cpp
```diff
@@ -29,7 +30,43 @@ create_link_layer(boost::asio::io_service& io_service, const EthernetDevice& dev
...
         link_layer.reset(new UdpLink { io_service, multicast });
+    } else if (name.substr(0, 3) == "tcp") {
+        namespace ip = boost::asio::ip;
+
+        TcpLink* tcp = new TcpLink { io_service };
+
+        std::string tcp_ip;
+        unsigned short tcp_port, tcp_ip_len;
+
+        if (vm.count("tcp-connect")) {
```
riebl, line 41: I am not sure if `vm.count` is necessary: An empty vector of string would be just fine.

tools/socktap/tcp_link.hpp
```diff
@@ -0,0 +1,64 @@
...
+    ~TcpLink();
+
+    void indicate(IndicationCallback) override;
+    void request(const vanetza::access::DataRequest&, std::unique_ptr<vanetza::ChunkPacket>) override;
+    void connect(boost::asio::ip::tcp::endpoint);
+    void accept(boost::asio::ip::tcp::endpoint);
+    void accept_handler(boost::system::error_code& ec, TcpSocket* ts, boost::asio::ip::tcp::endpoint ep);
+
+private:
+    std::list<TcpSocket*> sockets_;
```
riebl, line 54: using `std::list<TcpSocket>` would eliminate the need for manual memory management with `new` and `delete`

tools/socktap/tcp_link.hpp
```diff
@@ -0,0 +1,64 @@
...
+    std::map<boost::asio::ip::tcp::endpoint, boost::asio::ip::tcp::acceptor*> acceptors_;
```
riebl, line 55: Same here, storing `acceptor` objects instead of pointers simplifies memory management.

tools/socktap/tcp_link.cpp
```diff
@@ -0,0 +1,169 @@
+#include "tcp_link.hpp"
+#include <vanetza/access/data_request.hpp>
+#include <vanetza/net/ethernet_header.hpp>
+#include <boost/asio/ip/multicast.hpp>
```
riebl, line 4: You are not using any multicast feature for TCP, are you?

tools/socktap/tcp_link.cpp
```diff
@@ -0,0 +1,169 @@
...
+                is_connected_ = true;
+                ByteBuffer buffer(rx_buffer_.begin(), rx_buffer_.begin() + length);
+                CohesivePacket packet(std::move(buffer), OsiLayer::Link);
+                if (packet.size(OsiLayer::Link) < EthernetHeader::length_bytes) {
+                    std::cerr << "Dropped TCP packet too short to contain Ethernet header\n";
+                } else {
+                    packet.set_boundary(OsiLayer::Link, EthernetHeader::length_bytes);
+                    auto link_range = packet[OsiLayer::Link];
+                    EthernetHeader eth = decode_ethernet_header(link_range.begin(), link_range.end());
+                    if (callback_) {
```
riebl, line 141: You need to check if the callback is set before invoking it: `if (*callback_)`

riebl (commented):
I like the addition of a TCP link layer, this is cool stuff!
Please have a look at my comments which I would like to sort out before merging. Thanks!

riebl:
Nice! Let me give it a final try, I will merge it then.

fysch:
I've just encountered some issues with it today.
I will try to resolve them in the next days.
So, don't worry about testing right now.

tools/socktap/tcp_link.cpp
```diff
@@ -0,0 +1,160 @@
...
+        status_ = CONNECTED;
+        do_receive();
+    }
+
+}
+
+void TcpSocket::request(std::array<boost::asio::const_buffer, layers_> const_buffers)
+{
+    boost::system::error_code ec;
+    socket_.write_some(const_buffers, ec);
```
riebl, line 120: `write_some` may only transmit the buffers incompletely. I think you should use `write` here.

tools/socktap/tcp_link.cpp
```diff
@@ -0,0 +1,160 @@
...
+            &TcpSocket::receive_handler,
+            this,
+            boost::placeholders::_1,
+            boost::placeholders::_2
+        )
+    );
+}
+
+void TcpSocket::receive_handler(boost::system::error_code ec, std::size_t length) {
+    if (!ec) {
```
riebl, line 141: TCP streams have no message boundaries by design. Thus, we need to ensure that we take full packets out of the stream. One possibility would be a size field (of fixed size) at the start of each packet. The receiver can read as many bytes as indicated by this field and knows that the next read must be the size field for the follow-up packet.

Example: [uint16: 500] 500 bytes packet [uint16: 400] 400 bytes packet [uint16 ...] ...

riebl (changes requested):
I have just noticed that there is no frame delimiter in the TCP stream. Your implementation cannot cope with situations where the IP stack transmits/receives data in chunks of different sizes than your socket calls pass/expect.

riebl:
Thanks a lot for your patience while contributing your TCP link layer!

PR #199, opened 2023-09-14
Add CPM (TS102894-2v211) to Vanetza

diogopjesus:
This is a possible solution to add the new CPM format as described in TS102894-2v211.
"its2" might be changed to a more appropriate name.
This is a straightforward solution to keep multiple versions of the Common Data Dictionary (CDD) and Collective Perception Message (CPM), without changing the previously generated code from older versions.

### DIFF
--- a/tools/socktap/tcp_link.cpp
+++ b/tools/socktap/tcp_link.cpp
@@ -30,7 +30,7 @@ void TcpLink::request(const access::DataRequest& request, std::unique_ptr<ChunkP
     vanetza::ByteBuffer buffer = create_ethernet_header(request.destination_addr, request.source_addr, request.ether_type);
 
     // insert packet size before ethernet header
-    uint16_t packet_size = packet->size() + EthernetHeader::length_bytes;
+    uint16_t packet_size = packet->size(OsiLayer::Network, OsiLayer::Application) + EthernetHeader::length_bytes;
     buffer.insert(buffer.begin(), packet_size & 0x00FF);
     buffer.insert(buffer.begin(), (packet_size & 0xFF00) >> 8);
 


riebl:
I like the idea of having an "its2" ASN.1 target with all the message formats of ITS release 2. Unfortunately, the "classic" CAM and DENM are not yet updated, i.e. their latest release still refers to CDD 2 whereas the latest CDD is 3.1.
I had no closer look at the CDD changes yet, but I would expect them to be backward-compatible; otherwise, the whole ASN.1 story gets absurd.
Since one cannot mix "its" and "its2" in a single executable, I would like "its2" to behave as drop-in replacement of "its". At least CAMs and DENMs should be supported by "its2", ideally also MAPEM, SPATEM, RTCMEM etc.

diogopjesus:
Hi @riebl,
The problem with the CDD changes is that some fields with the same name are currently declared in multiple asn1 files, like the Basic Container on the [currently available CAM](https://github.com/diogopjesus/vanetza/blob/4701564694d28aa31925bf42f42c576918a4b409/asn1/EN302637-2v141-CAM.asn#L57) and on the [newest CDD](https://github.com/diogopjesus/vanetza/blob/4701564694d28aa31925bf42f42c576918a4b409/asn1/TS102894-2v211-CDD.asn#L4242). This fields would be renamed with a prefix attached to its original name when generating code from these two asn1 files to the same directory. We would have ```CAM-PDU-Descriptions_BasicContainer.h``` and ```ITS-Container_BasicContainer.h``` inside _its_.

riebl:
I have just noticed that you have also slightly patched the original CPM ASN.1 source, haven't you?

v0-e:
> Edit: I have tried the most recent version of @mouse07410's asn1c ([mouse07410/asn1c@75cc775](https://github.com/mouse07410/asn1c/commit/75cc775974d1e17bfd5efeaf1345180a1f45b8e6)) but it also fails with `ASN.1 grammar parse error near [...]/vanetza/asn1/TS103324v211-CPM-PDU-Descriptions.asn:11 (token "WITH"): syntax error`. This is quite a pity: Either patch the original ASN.1 sources just as @diogopjesus did or we need to enhance asn1c.

This [fork](https://github.com/fillabs/asn1c) seems to be able to parse the original CPM ASN.1 source.

riebl:
Thanks for your hint @v0-e, I will give it a try!

mouse07410:
@v0-e would you care to number a PR to add the missing capability here?

riebl:
See #209 for the follow-up PR.

PR #200, opened 2023-09-25
udp multicast ip

valt2017:
Specific network interface (selected according its IP address) is used for UDP multicast.

### DIFF
--- a/tools/socktap/udp_link.cpp
+++ b/tools/socktap/udp_link.cpp
@@ -7,18 +7,28 @@
 namespace ip = boost::asio::ip;
 using namespace vanetza;
 
-UdpLink::UdpLink(boost::asio::io_service& io_service, const ip::udp::endpoint& endpoint) :
+UdpLink::UdpLink(boost::asio::io_service& io_service, const ip::udp::endpoint& endpoint, const std::string& ip) :
     multicast_endpoint_(endpoint),
     tx_socket_(io_service), rx_socket_(io_service),
     rx_buffer_(2560, 0x00)
 {
     tx_socket_.open(multicast_endpoint_.protocol());
 
+    if(ip.size() != 0){
+        boost::asio::ip::multicast::outbound_interface option(boost::asio::ip::address_v4::from_string(ip));
+        tx_socket_.set_option(option);
+    }
+
     rx_socket_.open(multicast_endpoint_.protocol());
     rx_socket_.set_option(ip::udp::socket::reuse_address(true));
     rx_socket_.bind(multicast_endpoint_);
     rx_socket_.set_option(ip::multicast::enable_loopback(false));
-    rx_socket_.set_option(ip::multicast::join_group(multicast_endpoint_.address()));
+
+    if(ip.size() != 0){
+        rx_socket_.set_option(ip::multicast::join_group(multicast_endpoint_.address().to_v4(), boost::asio::ip::address_v4::from_string(ip)));
+    }else{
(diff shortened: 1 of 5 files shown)


tools/socktap/udp_link.cpp
```diff
@@ -7,18 +7,28 @@
 namespace ip = boost::asio::ip;
 using namespace vanetza;
 
-UdpLink::UdpLink(boost::asio::io_service& io_service, const ip::udp::endpoint& endpoint) :
+UdpLink::UdpLink(boost::asio::io_service& io_service, const ip::udp::endpoint& endpoint, const std::string& ip) :
```
riebl, line 10: passing address as `boost::asio::ip::address` might be beneficial

tools/socktap/udp_link.cpp
```diff
@@ -7,18 +7,28 @@
...
     multicast_endpoint_(endpoint),
     tx_socket_(io_service), rx_socket_(io_service),
     rx_buffer_(2560, 0x00)
 {
     tx_socket_.open(multicast_endpoint_.protocol());
 
+    if(ip.size() != 0){
+        boost::asio::ip::multicast::outbound_interface option(boost::asio::ip::address_v4::from_string(ip));
```
riebl, line 18: I just had a look at the *ip(7)* man page: I assume we can even pass the interface index instead of an IP address to `outbound_interface` and `join_group`. This would save us the `EthernetDevice::ip` lookup call.

valt2017: Interface index instead of IP seems reasonable, I can try it later.

riebl: This PR has been almost forgotten: Have you tried the approach using an interface index directly?

valt2017: I haven't tried it yet :-) I can play with it in pure C during the weekend. ip_mreqn structure seems like a good start according to the manual.

valt2017: In pure C, it's possible to use ip_mreqn structure instead of ip_mreq. Name of the interface and its index is enough to use specific interface for multicast.
 
  struct ip_mreqn mreq;
  mreq.imr_multiaddr.s_addr = inet_addr("239.1.1.1");
  mreq.imr_address.s_addr = htonl(INADDR_ANY);
  mreq.imr_ifindex = if_nametoindex("eth0"); 
...
setsockopt(sock, IPPROTO_IP, IP_MULTICAST_IF, &mreq, sizeof(mreq))
...
setsockopt(sock, IPPROTO_IP, IP_ADD_MEMBERSHIP, &mreq, sizeof(mreq))

In boost ASIO sources only ip_mreq type is used (../detail/socket_types.hpp):
...
typedef ip_mreq in4_mreq_type;

So some own wrapper should be created for socket options to use interface index?

riebl: On the one hand, a wrapper for this feature would be nice, on the other hand, I think it's not worth the effort for socktap.
Let's just stick with existing and documented `join_group` and `outbound_interface`.
I have created a follow-up commit to your PR at https://github.com/riebl/vanetza/tree/multicast.
Can you @valt2017 please cross-check it?

valt2017: I've tested your modifications and binding to IP address seems to be working as in previous version. When socket is switched to AF_INET type, obtaining the IP address using ioctl also works properly. :thumbsup:

riebl:
This PR is rebased and committed as 5dc7c543075c720cb048ec7c9f1ba4ea86a47907

Issue #154, opened 2022-04-21
compilation of ASN.1 files


liviolima80:
Hi, 
I'm trying to recompile the ASN.1 files with make generate_asn1c

I have an error when the patch its-circular-dependencies.patch is applied:

-- Applying patch its-circular-dependencies
error: vanetza/asn1/its/ConnectionManeuverAssist-addGrpC.c: No such file or directory
error: vanetza/asn1/its/ConnectionManeuverAssist-addGrpC.h: No such file or directory
error: vanetza/asn1/its/MapData-addGrpC.c: No such file or directory
error: vanetza/asn1/its/MapData-addGrpC.h: No such file or directory
error: vanetza/asn1/its/NodeSetXY.c: No such file or directory
error: vanetza/asn1/its/NodeSetXY.h: No such file or directory
CMake Error at vanetza/asn1/apply_patches.cmake:8 (message):

Any idea about the issue?

Regards

riebl:
Thanks for reporting this issue. Regenerating the ASN.1 sources was not possible without setting the CMake option `VANETZA_ASN1_WITH_ISO` to *ON*. This is fixed now.

liviolima80:
Hi, 
I fixed the issue adding VANETZA_ASN1_WITH_ISO flag, but in order to complete the cmake configuration a recent cmake version is needed. With 3.10 for example you get the following error:

CMake Error at vanetza/asn1/CMakeLists.txt:34 (list):
  list does not recognize sub-command TRANSFORM

I updated to 3.20 and the error is fixed

riebl:
Thanks for this hint; I have updated the minimum required version of CMake to 3.12, which introduced `list(TRANSFORM …)`.

PR #202, opened 2023-10-19
Regenerating the ASN.1 sources was not possible without setting the CMake option VANETZA_ASN1_WITH_ISO to ON

khevessy:
Fix of regression to #154


riebl:
Thank you!

Issue #181, opened 2023-01-27
Crash from manual runtime scheduling of Received CAM messages


sivasr1:
Hello everyone,

While we are trying to reproduce the socktap binary with CAM application for DEMO purpose we always get a crash over time from
the function "ManualRuntime::trigger(Clock::time_point tp)" defined in "vanetza/common/manual_runtime.cpp"
This crash (mainly on reception side and rarely on transmission side) happens only after successful processing of many received CAM messages.

From the debugging the main reason for this crash seems to be because of the simultaneous calls to function
"void TimeTrigger::schedule()" defined in "tools/socktap/time_trigger.cpp"

NOTE :: The crash issue is always reproducible , but crash happens quicker if the rate of transmitted messages is high  ( example :: 100 milliseconds/1 CAM message). So basically receiving 10 messages in a second runs about 40-60 minutes without any issues but then this crash happens. Once we reduce the rate of the receival of the messages from 10 to 1 in a second then actually our system runs for multiple hours before having the same crash.

Please refer commands below for more information related to GDB back trace and socktap commands usage.

Core issue::
----------------
socktap: vanetza/common/manual_runtime.cpp:62: void vanetza::ManualRuntime::trigger(vanetza::Clock::time_point): Assertion `tp >= m_now' failed.
8772 64 
Aborted (core dumped)

On Transmission side::
----------------------

./socktap -a ca -p static --latitude 40.8810 --longitude 29.3433 --pos_confidence 5 --print-tx-cam --cam-interval 100 -i lo --mac-address 00:00:00:00:00:11

On Reception Side::
--------------------
./socktap -a ca -p static --latitude 40.8810 --longitude 29.3433 --pos_confidence 5 --print-rx-cam -i lo --mac-address 00:00:00:00:00:22 

GDB backtrace::
----------------
(gdb) bt
#0  0x0000ffff94f789a8 in raise () from /lib/libc.so.6
#1  0x0000ffff94f65f54 in abort () from /lib/libc.so.6
#2  0x0000ffff94f72148 in ?? () from /lib/libc.so.6
#3  0x0000ffff94f721ac in __assert_fail () from /lib/libc.so.6
#4  0x0000aaaac557a868 in vanetza::ManualRuntime::trigger (this=this@entry=0xffffee3a61e0, tp=...)
    at /opt/fsl-imx-wayland/5.4-zeus/sysroots/aarch64-poky-linux/usr/include/c++/9.2.0/bits/locale_facets.h:877
[...]

PR #205, opened 2023-11-07
Use boost::asio::post in asynchronous callback in Autotalks link layer

khevessy:
I noticed that in Autotalks link layer, sometimes we get the same error as mentioned [here](https://github.com/riebl/vanetza/issues/181#issue-1559609353), so I tried to use `boost::asio::post` as @riebl suggested. This seems to fix the issue but I do not know if it is 100% correct.


riebl:
Looks good to me, though I cannot test it without Autotalks device.

PR #206, opened 2023-11-08
socktap: fix packet size in TcpLink

fysch:
When the router forwards a packet, the packet is duplicated and therefore a link layer is already present.
In this case, the packet-size calculation would count the link layer twice.


tools/socktap/tcp_link.cpp
```diff
@@ -31,6 +31,9 @@ void TcpLink::request(const access::DataRequest& request, std::unique_ptr<ChunkP
 
     // insert packet size before ethernet header
     uint16_t packet_size = packet->size() + EthernetHeader::length_bytes;
```
riebl, line 33: I think the fix should be replacing `packet->size()` by `packet->size(OsiLayer::Network, OsiLayer::Application)`.

fysch: Yes, that is a more elegant solution. I didn't know about this function.

riebl:
Thanks, I have squashed your PR and merged it per 711e424e.

PR #207, opened 2023-11-13
socktap: remove gpsd status field from checks

firolino:
The gpsd status field is never set >= 0 despite mode is = 3D in my 3.25 installation. According to https://gitlab.com/gpsd/gpsd/-/issues/183 the status field was never meant to be used for validating the fix.


riebl:
Thanks a lot for your contribution!

PR #208, opened 2023-11-20
Add support for CUBE EVK

firolino:
Adds support for nfiniity's [CUBE EVK](https://www.nfiniity.com/#portfolio) into socktap.


riebl (changes requested):
Looks good! However, it adds an unnecessary dependency to Boost::coroutine.

riebl:
Thanks for your contribution!

PR #215, opened 2024-01-03
socktap: use protobuf::libprotobuf target

riebl:
Modern CMake prefers targets to depend on libraries and their include directories.


riebl:
Fixed in master.

PR #217, opened 2024-01-26
asn1: Add print() method to asn1c wrapper

v0-e:
Adds new `print()` method to ASN.1 types by leveraging asn1c's `asn_fprint()`. 
Defaults to `stdout` though the output stream can be specified. Specifying the output is done through a `FILE*`, since there is no straightforward way to convert between an `std::ostream` to a `FILE*`.
Can be regarded as an alternative to CAM's `print_indented()`.


riebl:
Thanks!

PR #218, opened 2024-01-29
asn1: Add compare methods to asn1c wrapper

v0-e:
Adds new `compare()` and related `==` and `!=` overloads to the asn1c wrapper.
Leverages asn1c's `compare_struct` operation, which tests the equality of two ASN.1 structs of the same type.

### DIFF
--- a/vanetza/asn1/tests/asn1c_wrapper.cpp
+++ b/vanetza/asn1/tests/asn1c_wrapper.cpp
@@ -68,11 +68,10 @@ TEST(asn1c_wrapper, compare) {
     OCTET_STRING_fromString(&wrapper4->string, "1234");
 
     // .compare()
-    EXPECT_TRUE(!wrapper1.compare(wrapper1));
-    EXPECT_TRUE(!wrapper1.compare(wrapper2));
-    EXPECT_FALSE(!wrapper1.compare(wrapper3));
-    EXPECT_TRUE(wrapper1.compare(wrapper3) > 0);
-    EXPECT_FALSE(!wrapper1.compare(wrapper4));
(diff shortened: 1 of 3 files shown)


vanetza/asn1/tests/asn1c_wrapper.cpp
```diff
@@ -56,6 +56,32 @@ TEST(asn1c_wrapper, validate) {
...
+    test_wrapper wrapper2(asn_DEF_VanetzaTest);
+    OCTET_STRING_fromString(&wrapper2->string, "1234");
+    test_wrapper wrapper3(asn_DEF_VanetzaTest);
+    OCTET_STRING_fromString(&wrapper3->string, "0123");
+    test_wrapper wrapper4(asn_DEF_VanetzaTest);
+    wrapper4->field = 5;
+    OCTET_STRING_fromString(&wrapper4->string, "1234");
+
+    // .compare()
+    EXPECT_TRUE(!wrapper1.compare(wrapper1));
```
riebl, line 71: I think the tests should check the actual integer value here.

v0-e: Thanks for the input. Do you mean add additional `EXPECT_*`s for wrappers with only the integer field populated?

riebl: The newly added `compare` method returns an integer but the tests here treat the return value only as a boolean.
I suggest to replace `EXPECT_TRUE` by `EXPECT_EQ(wrapper1.compare(wrapper1), 0)` etc.

riebl:
Thanks for your contribution @v0-e!

PR #219, opened 2024-01-30
asn1: OER-based copy for asn1c wrapper

v0-e:
May I suggest the use of OER to do the roundtrip-based copy of ASN.1 types?
Unless there are any concerns with the implementation/safety of asn1c's OER, we can consider it over the current BER/DER copy since it is much faster.

A benchmark of 1000000 roundrip encoding/decodings of a zero-initialized CAM with some fields set to `1` to pass the constraint checks:
```
OER Roundtrip in 3.370587 s   // (canonical)
BER Roundtrip in 14.393260 s  // der -> ber
PER Roundtrip in 3.833098 s   // (canonical unaligned)
JER Roundtrip in 10.260315 s  // (minified)
XER Roundtrip in 14.103931 s  // (canonical)
```

### DIFF
--- a/vanetza/asn1/asn1c_wrapper.cpp
+++ b/vanetza/asn1/asn1c_wrapper.cpp
@@ -52,16 +52,16 @@ void* copy(asn_TYPE_descriptor_t& td, const void* original)
     ByteBuffer buffer;
 
     asn_enc_rval_t ec;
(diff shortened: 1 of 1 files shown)


riebl:
Fine for me :-)

PR #223, opened 2024-03-18
Update ETSI TS 103 097 & 102 941 security specs

DeeJayBro:
Updates the security-related ASN definitions for TS 102 941 to v2.2.1 and TS 103 097 to v2.1.1.

As the IEEE definitions need some patching and with the introduction of IEEE1609.2.1 in TS 102 941 plus the resulting increase in the amount of files, these were outsourced to a cmake script which also takes care of the patching

Compiling to code needs an asn1c with the following pull request (https://github.com/mouse07410/asn1c/pull/138), as newer ETSI ITS ASN.1 defs make great use of `WITH SUCCESSORS`


vanetza/asn1/fetch_ieee_asn1_files.cmake
```diff
@@ -0,0 +1,116 @@
+if(NOT DIRECTORY)
+    message(FATAL_ERROR "DIRECTORY variable is undefined")
+endif()
+file(MAKE_DIRECTORY "${DIRECTORY}")
+
+file(DOWNLOAD
+    https://forge.etsi.org/rep/ITS/asn1/ieee1609.2/-/raw/2022-published/Ieee1609Dot2.asn
+    "${DIRECTORY}/IEEE1609dot2.asn")
+file(READ "${DIRECTORY}/IEEE1609dot2.asn" _content)
+string(REGEX REPLACE " Extension" " IEEEExtension" _content "${_content}")
```
riebl, line 10: Wow, I-triple-E becomes I-quadruple-E. Maybe go for `IeeeExtension`? Looks more readable in my opinion.

vanetza/asn1/fetch_ieee_asn1_files.cmake
```diff
@@ -0,0 +1,116 @@
...
+* @brief This structure defines the all certificate request structures as a
+* scoped version of the ScmsPdu.
+*/
+ScopedCertificateRequest ::= ScmsPdu (
+ScmsPdu-RaAcaCertRequest | ScmsPdu-EeEcaCertRequest | 
+ScmsPdu-EeRaCertRequest  | ScmsPdu-EeRaSuccessorEnrollmentCertRequest
+)
+/**
+ * @brief This structure defines the format of a signed certificate request.")
+string(REGEX REPLACE "ScopedCertificateRequest ::=.+signed certificate request." "${_valid_scoped_certificate_request}" _content "${_content}")
```
riebl, line 115: It becomes quite hard to see the actual patch this way. I prefer to have a patch file then. Is there a reason why a patch file is inappropriate here?

DeeJayBro: No, the complexity only started to explode after beginning that script already so i didn't change it to a patch, but good idea!

DeeJayBro:
Any update on this? should the asn1 definitions moved to `release2` to allow v2.1.1 and v1.3.1 to coexist?

DeeJayBro:
I just rechecked, replacing the "old" definitions should also be workable, since the standard extends a few datatypes and moves over to directly to IEEE 1609 but not changing on outer-facing definitions, with all tests still passing on a slate using the new asn.1 files

riebl:
Thanks for your contribution @DeeJayBro!

Issue #225, opened 2024-05-03
Parsing Secured Packet v3 with Certificate in signer container


Rom-1T:
Hi everyone,

I'm working on implementing Security v3 using Vanetza. I'm encountering an issue when attempting to use a certificate within the signer container. I'm receiving the error: "packet because of Parse_Secured_Header (2)".

I initially suspected the issue might be related to the ASN.1 library version. However, upgrading to version 2.2.1 using the pull request https://github.com/riebl/vanetza/pull/223/commits did not resolve the problem.

Has anyone else tried to decode a Secured Packet v3 that includes a certificate inside the signer container? Any insights or experiences you can share would be greatly appreciated.

valt2017:
Is your modified code available somewhere? Otherwise, I'm not sure if anyone will be able to help.

riebl:
Alternatively, a dump of the packet for which the parsing fails may be helpful.

Rom-1T:
[...]

PR #226, opened 2024-05-22
Bug Fix : OER constraint for asn1/its/Elevation.c + UPER contraints for ans1/its/DeltaReferencePositions & ans1/its/DeltaPositions

Rom-1T:
Hello,

After this issue : https://github.com/riebl/vanetza/issues/225. I've come accross to a bug, the Elevation.c in asn/its was causing a problem for the decoding of secured packet. I've just changed the OER constraint to limit potential repercussion outside of secured packet.

Also, I've discovered that when I receives IVIM packets with PolygonalLine/DeltaPositions_t I have problem decoding. It appears that the UPER constraint was wrong, I've only looked at DeltaPositions and DeltaReferencePositions.


vanetza/asn1/its/DeltaPositions.c
```diff
@@ -15,7 +15,7 @@ static asn_oer_constraints_t asn_OER_type_DeltaPositions_constr_1 CC_NOTUSED = {
 #if !defined(ASN_DISABLE_UPER_SUPPORT) || !defined(ASN_DISABLE_APER_SUPPORT)
 asn_per_constraints_t asn_PER_type_DeltaPositions_constr_1 CC_NOTUSED = {
 	{ APC_UNCONSTRAINED,	-1, -1,  0,  0 },
-	{ APC_CONSTRAINED | APC_EXTENSIBLE,  7,  7,  1,  100 }	/* (SIZE(1..100,...)) */,
+	{ APC_CONSTRAINED | APC_EXTENSIBLE,  5,  5,  1,  32 }	/* (SIZE(1..100,...)) */,
```
riebl, line 18: We should not patch the generated sources but update the used ISO 19321 ASN.1 file which includes the changed sequence constraints. At the moment, we fetch "ed-2" in fetch_iso_asn1_files.cmake, the changed constraints are included in "ed-3".

vanetza/asn1/its/DeltaReferencePositions.c
```diff
@@ -15,7 +15,7 @@ static asn_oer_constraints_t asn_OER_type_DeltaReferencePositions_constr_1 CC_NO
 #if !defined(ASN_DISABLE_UPER_SUPPORT) || !defined(ASN_DISABLE_APER_SUPPORT)
 asn_per_constraints_t asn_PER_type_DeltaReferencePositions_constr_1 CC_NOTUSED = {
 	{ APC_UNCONSTRAINED,	-1, -1,  0,  0 },
-	{ APC_CONSTRAINED | APC_EXTENSIBLE,  7,  7,  1,  100 }	/* (SIZE(1..100,...)) */,
+	{ APC_CONSTRAINED | APC_EXTENSIBLE,  5,  5,  1,  32 }	/* (SIZE(1..32,...))*/,
```
riebl, line 18: Same here, update ISO 19321 ASN.1 used for code generation instead.

vanetza/asn1/its/Elevation.c
```diff
@@ -38,7 +38,7 @@ Elevation_constraint(const asn_TYPE_descriptor_t *td, const void *sptr,
  */
 #if !defined(ASN_DISABLE_OER_SUPPORT)
 static asn_oer_constraints_t asn_OER_type_Elevation_constr_1 CC_NOTUSED = {
-	{ 4, 0 }	/* (-4096..61439) */,
+	{ 2, 1 }	/* (-4096..61439) */,
```
riebl, line 41: I see that the number of values fits in an `uint16_t`. Does asn1c already shift the values as expected, i.e. a raw 0 is decoded as -4096 in `Elevation_t`?

riebl: Section 2.2 of https://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf summarizes the OER encoding rules for integer types. According to this summary, the Elevation type with a lower bound of -4096 and an upper bound of 61439 is encoded as a signed integer in four octets.

PR #227, opened 2024-05-24
Fuzz harness

riebl:
@Boslx has kindly provided a harness for fuzz testing based on AFL++.
I have contributed some scripts for conveniently executing these tests in a Docker environment.


riebl:
Received positive feedback from @Boslx via e-mail. Merging.

PR #230, opened 2024-07-01
Security 2.1.1 - Tx and RX

Rom-1T:
Hello,

For my project which uses Vanetza, I needed to implement secure message sending. I've implemented the DummySignService and the StraightSignService. While I don't think it's ready yet for integration into the master branch, I would appreciate your feedback on this implementation. I've tested it using socketap and it seems to be working fine.

Thank you!


riebl:
Looks good after a first quick review :-)
You may have noted that working with these v3 data structures generated by asn1c is little fun.

Rom-1T:
Yes, it was sometimes tricky with all the pointer declarations, but that is what makes it interesting. If it needs any improvement whatsoever, please let me know.

riebl:
I have rebased your PR and merged it into the master branch now. Thanks for your contribution!

PR #237, opened 2024-11-12
Add InterfaceType LTE_V2X

khevessy:
ETSI EN 302 636-4-1 v1.4.1 has three itsGnIfType - Unspecified, its-g5 and lte-v2x.


riebl:
Thanks Karel!

PR #241, opened 2024-12-13
Small fixes

khevessy:
Hi,
couple of small changes:
1. Pretty sure by this - `vanetza::security::v3::save_certificate_to_file()` has `certificate` parameter `const` in header but not `const` in implementation. Adding it there requires that `vanetza::security::v3::serialize()` for certificate has this parameter also `const`. Further, small comment fix.
2. Small cleanup - remove `<future>` header inclusion from `vanetza/security/v2/sign_service.cpp` and `vanetza/security/v2/signature.hpp` as it seems not needed here, just in `vanetza/security/signature.hpp`. This was just for my convenience when building library for C++/CLI wrapper, since that does not support `<future>` and I wanted it in as few places as possible. Please ignore this if you think it is not needed.

### DIFF
--- a/vanetza/security/v2/sign_service.cpp
+++ b/vanetza/security/v2/sign_service.cpp
@@ -5,6 +5,7 @@
 #include <vanetza/security/v2/sign_header_policy.hpp>
 #include <vanetza/security/v2/sign_service.hpp>
 #include <vanetza/security/v2/signature.hpp>
+#include <future>
 
 namespace vanetza
 {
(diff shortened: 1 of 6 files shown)


vanetza/security/v2/sign_service.cpp
```diff
@@ -5,7 +5,6 @@
 #include <vanetza/security/v2/sign_header_policy.hpp>
 #include <vanetza/security/v2/sign_service.hpp>
 #include <vanetza/security/v2/signature.hpp>
-#include <future>
```
riebl, line 8: <future> also brings `std::async` used in line 76

riebl:
An almost perfect PR :-)
Please leave `#include <future>` in for `v2::DeferredSignService`.

khevessy:
I was trying really hard :D
Hope now it is ok.

PR #243, opened 2025-01-20
Removing Autotalks find operation in Root CMakeLists.txt

buraksenkus:
Since the Autotalks library is used for only **socktap** tool and is optional, the find operation should be removed from the root `CMakeLists.txt`. I think, it is actually the same case with Cohda libraries' find operation.


riebl:
Thank you!

Issue #245, opened 2025-01-28
TimestampIts value too large error on 32bit machine


buraksenkus:
I'm trying to initialize and validate a `vanetza::asn1::Denm` object, but I cannot do so on 32-bit (MK5 armhf) machines. I encounter the error:
`TimestampIts: value too large (/facilities/extern/vanetza/vanetza/asn1/its/TimestampIts.c:26)`

To narrow down the scope of the issue, I updated `tests/its.cpp` as follows and ran it on the MK5. However, the test did not pass.

```
TEST(ItsAsn1, encode_denm)
{
    asn1::Denm denm;
    EXPECT_EQ(0, asn_uint642INTEGER(&denm->denm.management.detectionTime, 4398046511103));  // Used max valid value
    EXPECT_EQ(0, asn_uint642INTEGER(&denm->denm.management.referenceTime, 4398046511103));  // Used max valid value
    EXPECT_TRUE(denm.validate());
    vanetza::ByteBuffer buf = denm.encode();
    EXPECT_EQ(40, buf.size());
}
```

```
[ RUN      ] ItsAsn1.encode_denm
/facilities/extern/vanetza/vanetza/asn1/tests/its.cpp:30: Failure
Value of: denm.validate()
  Actual: false
Expected: true
[  FAILED  ] ItsAsn1.encode_denm (1 ms)
```

I found that issues #78 and #138 describe similar situations, and they seem to have been resolved. Additionally, we had a working artifact a couple of years ago. However, the latest version appears to be faulty.

To investigate further, I reviewed previous commits to identify which one introduced the issue. The first problematic commit seems to be [57e37984a4e481de9797e30de679417904c7f511](https://github.com/riebl/vanetza/commit/57e37984a4e481de9797e30de679417904c7f511). Do you have any idea about the problem?

Moreover, I suggest updating the `encode_denm` test as shown above since using the maximum valid values tends to be a more challenging and meaningful case for unit testing compared to using `0`. And, I'm wondering if there is a way to integrate 32-bit arm testing into CI pipeline? Is this a feature that you would like in the pipeline?

khevessy:
I do not know how to properly fix this, I just wanted to say that I have the same problem on Autotalks Craton 2 (also 32-bit ARM), as `long` is only 4-byte so I think that the check for bounds in `asn_INTEGER2long` in `TimestampIts_constraint()` fails.

Not sure how to fix that properly, for our usage I just changed manually `long` to `uint64_t`
[...]

PR #247, opened 2025-02-06
asn1: patch ITS timestamp for 32-bit targets (see #245)

khevessy:
Tried to implement aforementioned fix based on https://github.com/riebl/vanetza/blob/master/vanetza/asn1/patches/security-64bit-types.patch - as advised by @riebl in #245.

Raphael, please, I am not sure what does the `asn_INTEGER_specifics_t` structure mean, does it also have to be added here?


riebl:
Thanks for your work, Karel! Yes, I think the the `asn_INTEGER_specifics_t` structure is necessary as the encoding and decoding otherwise defaults to "signed integer", see https://github.com/riebl/vanetza/blob/ba4d1fb5b79d44f0c61dd7d13db0927586c577b0/vanetza/asn1/support/INTEGER_uper.c#L61
So from all these "specifics", the `field_unsigned` flag is important for us in this case.

khevessy:
Thanks for the explanation. Added the structure now.

PR #248, opened 2025-02-19
VAM v2.2.1

khevessy:
Hi, I tried to add VAM messages (version 2.2.1 - Release 2), since they were not added in [commit with Release 2 messages](https://github.com/riebl/vanetza/commit/3267b4efbfb0f69d43aceb388ebb20b06c191241). I omitted original VAM (version 2.1.1 from [this branch](https://github.com/khevessy/vanetza/tree/vam)), as since 4 years this is implemented, no one seems to have needed it. Maybe these could be added as Release 1 version (similar to CPM - the older version that we call non-Release 2 is per [ETSI TR 103 562 V2.1.1](https://www.etsi.org/deliver/etsi_tr/103500_103599/103562/02.01.01_60/tr_103562v020101p.pdf) which calls itself Release 2).


khevessy:
PS. https://forge.etsi.org/rep/ITS/asn1/vam-ts103300_3/-/blob/v.2.2.1/VAM-PDU-Descriptions.asn has a typo in the download URL in the header, I left it there since the actual correct URL is in the README in the folder.

riebl:
Thanks for your PR @khevessy! Which asn1c version have you used for generating the code? I am just wondering about some changes in existing files.

khevessy:
I am using 18e565032e52af8002c2353be20bdbba9233e700 as always which I think is also the version you are using. If you speak about the removal of "static" and the consequent extern exports, cannot it be because the compiler detected, that some structures are now needed for encoding / decoding of the messages? In fact, when I remove the newly added files from CMakeLists, these strange changes go away, so I think it cannot be caused by wrong asn1c version.

On another note, I deliberately leave out are three occurrences of EXTERNAL.h/EXTERNAL.c changes, because there is just a change in path in the header comment (on line 4 from `/home/raphael/opt/asn1c-18e56503` to `/usr/local/`) https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/its/EXTERNAL.c#L4

https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/its/EXTERNAL.h#L4

https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/its/r2/EXTERNAL.c#L4

https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/its/r2/EXTERNAL.h#L4

https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/security/EXTERNAL.c#L4

https://github.com/riebl/vanetza/blob/0b241f31540de482e1fd9f575e8b024ec7afbd0d/vanetza/asn1/security/EXTERNAL.h#L4

riebl:
Thanks for providing these details! Then let's go for it and merge it :-)

PR #254, opened 2025-05-02
asn1: add missing patch for Release 2 64bit types

khevessy:
I noticed that I have some compilation warning and I think that is by missing patch to Release 2 64bit types.


riebl:
Thanks!

PR #279, opened 2026-03-23
asn1: add rtcmem wrapper

buraksenkus:
I've added missing asn1c wrapper for RTCMEM

### DIFF
--- a/vanetza/asn1/rtcmem.hpp
+++ b/vanetza/asn1/rtcmem.hpp
@@ -1,0 +1,23 @@
+#ifndef RTCMEM_HPP_XAA1CDFFI
+#define RTCMEM_HPP_XAA1CDFFI
+
+#include <vanetza/asn1/asn1c_conversion.hpp>
+#include <vanetza/asn1/asn1c_wrapper.hpp>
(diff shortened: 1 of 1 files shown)


riebl:
Hi @buraksenkus,

I will add the RTCMEM wrapper. However, I am not a fan of adding the recursive mutexes because they will affect everyone including those running Vanetza single-threaded.
My usual pattern when integrating Vanetza in a multi-threaded application: All Vanetza entities (including security, DCC, access interface etc.) are running on a single thread and thread-safe queues are used for sharing incoming/outgoing packets with other application code potentially running on other threads.
If you have an idea for another (optional) synchronization mechanism, please create a separate pull request. Thanks!

buraksenkus:
Actually those commits shouldn't be in this branch, they were added by mistake. We're working on multi-thread operations in another branch and they are not related with this PR. Sorry for the mistake. I'll fix the PR

buraksenkus:
Now this is the intended PR :)

riebl:
Merged as 361409caff49115def6d5370ae0bae26bef7c4e4

PR #280, opened 2026-04-11
fix(v3/location_checker): permissive fallback for IdentifiedRegion

### LINKED ISSUE #262: security v3: valid_at_location
I am not completely sure if we did not already discuss this, but some of the location checks for V3 certificate are not implemented:

https://github.com/riebl/vanetza/blame/40776241ecc758100b0b3b17caac7dcdfcacb0ec/vanetza/security/v3/certificate.cpp#L107

https://github.com/riebl/vanetza/blame/40776241ecc758100b0b3b17caac7dcdfcacb0ec/vanetza/security/v3/certificate.cpp#L110

We encountered the Region one in real life. Our fix was to just put `return true;` in the appropriate case. To be honest this approach does not seem right, but basically silently failing took a little time to debug. Maybe there could be some option for toggling this check on/off?

Btw the region restriction seems like a nightmare to check based on location.

aki1770-del:
## Summary

`DefaultLocationChecker::valid_at_location()` returned `false` for any certificate carrying an `IdentifiedRegion` (ISO 3166-1 country codes or sub-regional identifiers). This caused `Verdict::OutsideRegion` for all V2X peers whose certificates encode country-level geographic restrictions — a silent rejection that effectively breaks V2X communication for those peers.

Closes #262

## Fix

Return `true` (permissive fallback) for `IdentifiedRegion` until an external country-boundary dataset is integrated. Silently rejecting every certificate that asserts a country-level restriction is worse than deferring the restriction check — the majority of deployed ETSI ITS networks issue certificates with `IdentifiedRegion` constraints.

A comment in the code documents the intended full implementation path (Boost.Geometry R-tree + OpenStreetMap admin-level polygon data), consistent with the approach documented in issue #262.

Also fixed:
- Removed unreachable `return false` after the `polygonalRegion` case (dead code — never executed since `is_inside()` was returned on the line above)
- Removed redundant `return true` after the switch block (control never reached it)

## Existing behavior audit (OPS-RULE-016)

- `location_checker.cpp:35-37` — `PR_identifiedRegion` case: comment "not supported yet", returns `false`
- `certificate_validator.cpp:31-32` — `valid_at_location() == false` → `Verdict::OutsideRegion`
- `location_checker.cpp:33-34` — `PR_polygonalRegion` case: dead `return false` unreachable after `is_inside()` return

## Scope classification (OPS-RULE-017)

**(b) Minor enhancement** — adds permissive fallback for an unimplemented region type; no change to other region type handling.

## Prior art consulted (OPS-RULE-018)

1. Issue [#262](https://github.com/riebl/vanetza/issues/262) — root cause description and discussion of the intended Boost.Geometry + boundary data approach
2. `vanetza/security/v3/geometry.hpp` + `boost_geometry.cpp` — existing Boost.Geometry integration for polygonal regions; the same infrastructure would be extended for IdentifiedRegion in a follow-up
3. `vanetza/security/v3/tests/default_certificate_validator_v3.cpp` — existing test suite pattern followed for the new `IdentifiedRegion` test case

## Test evidence

Added test case to `default_certificate_validator_v3.cpp` `region_validator` suite:
- Sets `GeographicRegion_PR_identifiedRegion` on a certificate
- Asserts `Verdict::Valid` (permissive fallback active)

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/tools/socktap/security.cpp
+++ b/tools/socktap/security.cpp
@@ -80,12 +80,14 @@ public:
 class SecurityContextV3 : public security::SecurityEntity
 {
 public:
-    SecurityContextV3(const Runtime& runtime, PositionProvider& positioning) :
+    SecurityContextV3(const Runtime& runtime, PositionProvider& positioning,
+                      bool permissive_identified_region = false) :
         runtime(runtime), positioning(positioning),
         backend(security::create_backend("default"))
     {
         cert_validator.use_runtime(&runtime);
         cert_validator.use_position_provider(&positioning);
+        location_checker.set_permissive_identified_region(permissive_identified_region);
         cert_validator.use_location_checker(&location_checker);
     }
 
@@ -212,7 +214,9 @@ create_security_entity(const po::variables_map& vm, const Runtime& runtime, Posi
             }
 
             if (version == 3) {
-                auto context = std::make_unique<SecurityContextV3>(runtime, positioning);
+                bool permissive_ir = vm.count("security.permissive-identified-region") &&
+                                     vm["security.permissive-identified-region"].as<bool>();
+                auto context = std::make_unique<SecurityContextV3>(runtime, positioning, permissive_ir);
                 context->cert_provider = load_v3_certificates(cert_path, cert_key_path, chain_paths);
                 context->build_entity();
                 security = std::move(context);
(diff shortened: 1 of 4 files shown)


riebl:
Hi @aki1770-del,

I am reluctant to silently accept any `IdentifiedRegion` without the user's consent (ITS station operator). However, I can imagine to add a flag changing the behaviour of `DefaultLocationChecker`.

Without actually implementing proper `IdentifiedRegion` checks, issue #262 cannot be closed.

Why do you think that "the majority of deployed ETSI ITS networks" uses `IdentifiedRegion`? So far, I have never encountered this restriction for any certificate I have received from an actually deployed station.

khevessy:
Hello, just my two cents to the discussion, we encountered the `IdentifiedRegion` present when working with certificates from Microsec PKI.
Karel

riebl:
Thanks!

PR #281, opened 2026-05-16
correct invalid warning directive in hmac.cpp

DeeJayBro:
probably no-one hit that edge-case before. but "warn" is not valid and thus leads the build to an fail and not a warning


riebl:
Thanks! Yes, I have obviously never hit this edge case :-)

PR #284, opened 2026-06-08
socktap: add selectable crypto backend

Abderraouf-D:


tools/socktap/security.cpp
```diff
@@ -266,6 +267,7 @@ void add_security_options(po::options_description& options)
 {
     options.add_options()
         ("security", po::value<std::string>()->default_value("dummy"), "Security entity [none,dummy,certs] (with optional -v2 or -v3 suffix)")
+        ("backend", po::value<std::string>()->default_value("default"), "Crypto backend [default,OpenSSL,CryptoPP,Null]")
```
riebl: A CLI option "--backend" might be too generic in the context of socktap. Maybe just "--crypto"? Happy to hear your thoughts!

Abderraouf-D: Good point.

What do you think about --crypto-backend? It is a bit longer than --crypto, but it is more explicit. If you think it is too long, then I am happy to go with your suggestion.

riebl: Fine for me!

PR #285, opened 2026-06-08
pki/certify: print subject keys and HTTP error details

mikelngelous:
TODO done:
- **certify** - show-certificate: decode and print the `Verification_Key`, `Encryption_Key` and `Reconstruction_Value` subject attributes.

Diagnostics:
- **pki**: include the HTTP status and response body in EA/AA/CPOC error messages.

### DIFF
--- a/tools/pki/authorization.cpp
+++ b/tools/pki/authorization.cpp
@@ -263,10 +263,8 @@ void request_one_ticket(Context& ctx, const Certificate& ec, const AuthoritiesFr
     auto response = http_post(query, "application/x-its-request", encoded);
 
     if (response.result() != boost::beast::http::status::ok) {
-        const auto& body = response.body();
-        std::string body_str(reinterpret_cast<const char*>(body.data()), body.size());
-        throw HttpException("AA returned an unexpected HTTP status for the authorization request: "
-            + std::to_string(response.result_int()) + " (" + body_str + ")");
+        throw HttpException("AA returned an unexpected HTTP status for the authorization request",
+            std::move(response));
     } else if (response[boost::beast::http::field::content_type] != "application/x-its-response") {
         throw HttpException("expected application/x-its-response");
     }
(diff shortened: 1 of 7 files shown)


tools/pki/authorization.cpp
```diff
@@ -263,7 +263,10 @@ void request_one_ticket(Context& ctx, const Certificate& ec, const AuthoritiesFr
     auto response = http_post(query, "application/x-its-request", encoded);
 
     if (response.result() != boost::beast::http::status::ok) {
-        throw HttpException("AA returned an unexpected HTTP status for the authorization request");
+        const auto& body = response.body();
+        std::string body_str(reinterpret_cast<const char*>(body.data()), body.size());
+        throw HttpException("AA returned an unexpected HTTP status for the authorization request: "
```
riebl: I don't like creating the error messages for exceptions dynamically, however, I admit that more context might be valuable.
My suggestion: Let's add the HTTP status code as parameter to `HttpException`. I am yet quite undecided if the body string shall be included in the exception, though.

mikelngelous: You're rght, Here's what I'd propose

- HttpException gets an extra constructor HttpException(what, status, body = {}) plus status() / body() accessors. The what message stays static at each call site.
- The status/body are formatted in a single place, the HttpException catch in main.cpp. So, no more dynamicall exception messages as you mention.
I'd lean towards keeping it as an optional constructor arg. When working with ETSI PKIs the status alone is often not enough to tell what went wrong, the reason is in the body (decrypt/signature/version errors, etc). But it's trivial to drop if you'd rather not,  the accessor just stays unused.

riebl: Might be even easier: `HttpException` optionally accepts the `HttpResponse` so we can move the status code and the body into the exception without requiring additional memory allocation.

mikelngelous: Done, http response moved into HttpException

riebl (approved):
Thanks!

PR #286, opened 2026-06-08
security: check v3 certificate chain consistency

Abderraouf-D:
This adds v3 certificate chain consistency checks between a subject certificate and its issuer.

The validator now checks:
- subject validity period is contained in issuer validity period
- requested ITS-AID is permitted by the issuer
- subject assurance does not exceed issuer assurance
- subject geographic region is contained in issuer region where supported

note: the permission consistency check is scoped to the requested ITS-AID, not all permissions in the AT. That is intentional because the current naive AT has IPV6_ROUTING, while the current AA does not issue that permission (we can add it later myabe).

### DIFF
--- a/vanetza/security/v3/certificate.cpp
+++ b/vanetza/security/v3/certificate.cpp
@@ -3,6 +3,7 @@
 #include <vanetza/security/v3/asn1_conversions.hpp>
 #include <vanetza/security/v3/certificate.hpp>
 #include <vanetza/security/v3/distance.hpp>
+#include <vanetza/security/v3/geometry.hpp>
 #include <boost/optional/optional.hpp>
 #include <cassert>
 #include <cstdlib>
@@ -28,6 +29,14 @@ bool is_compressed(const Vanetza_Security_EccP384CurvePoint& point);
 bool is_signature_x_only(const Vanetza_Security_Signature_t& sig);
 bool compress(Vanetza_Security_EccP256CurvePoint&);
 bool compress(Vanetza_Security_EccP384CurvePoint&);
+bool contains_permission(const Vanetza_Security_SequenceOfPsidSspRange_t& permissions, ItsAid aid);
+bool equal(const asn1::TwoDLocation& lhs, const asn1::TwoDLocation& rhs);
+bool equal(const asn1::RectangularRegion& lhs, const asn1::RectangularRegion& rhs);
+bool equal(const asn1::SequenceOfRectangularRegion& lhs, const asn1::SequenceOfRectangularRegion& rhs);
+bool is_within(const asn1::GeographicRegion& inner, const asn1::CircularRegion& outer);
+bool is_within(const asn1::GeographicRegion& inner, const asn1::SequenceOfRectangularRegion& outer);
+bool is_within(const asn1::GeographicRegion& inner, const asn1::PolygonalRegion& outer);
+bool is_within(const asn1::GeographicRegion& inner, const asn1::GeographicRegion& outer);
 bool make_x_only(Vanetza_Security_EccP256CurvePoint&);
 bool make_x_only(Vanetza_Security_EccP384CurvePoint&);
 bool make_signature_x_only(Vanetza_Security_Signature_t& sig);
@@ -142,6 +151,59 @@ bool valid_at_timepoint(const asn1::EtsiTs103097Certificate& cert, const Clock::
 bool CertificateView::valid_for_application(ItsAid aid) const
 {
     return m_cert ? v3::valid_for_application(*m_cert, aid) : false;
+}
+
+bool CertificateView::is_allowed_to_issue(ItsAid aid) const
+{
+    if (!m_cert) {
+        return false;
+    }
+
+    if (const auto* seq = m_cert->toBeSigned.certIssuePermissions; seq) {
+        for (int i = 0; i < seq->list.count; ++i) {
+            const auto* group = seq->list.array[i];
+            if (!group) {
+                continue;
+            }
+
+            const auto& subject_permissions = group->subjectPermissions;
+            if (subject_permissions.present == Vanetza_Security_SubjectPermissions_PR_all) {
+                return true;
+            } else if (
+                    subject_permissions.present == Vanetza_Security_SubjectPermissions_PR_explicit &&
+                    contains_permission(subject_permissions.choice.Explicit, aid)) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+
+boost::optional<std::uint8_t> CertificateView::assurance_level() const
+{
+    if (!m_cert || !m_cert->toBeSigned.assuranceLevel || m_cert->toBeSigned.assuranceLevel->size == 0) {
+        return boost::none;
+    }
+
+    return m_cert->toBeSigned.assuranceLevel->buf[0];
+}
+
+bool CertificateView::region_is_within(const CertificateView& issuer) const
+{
+    if (!m_cert || !issuer.m_cert) {
+        return false;
+    }
+
+    const auto* issuer_region = issuer.m_cert->toBeSigned.region;
+    const auto* subject_region = m_cert->toBeSigned.region;
+    if (!issuer_region) {
+        return true;
+    } else if (!subject_region) {
+        return false;
+    } else {
+        return is_within(*subject_region, *issuer_region);
+    }
 }
 
 bool valid_for_application(const asn1::EtsiTs103097Certificate& cert, ItsAid aid)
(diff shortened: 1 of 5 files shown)


vanetza/security/v3/certificate.hpp
```diff
@@ -238,6 +240,13 @@ boost::optional<Signature> get_signature(const asn1::EtsiTs103097Certificate& ce
  */
 std::list<ItsAid> get_aids(const asn1::EtsiTs103097Certificate& cert);
 
+/**
+ * Get list of ITS AID permissions a certificate is allowed to issue
+ * \param cert certificate
+ * \return list of ITS AIDs or empty list if all permissions are granted
+ */
+std::list<ItsAid> get_issuer_aids(const asn1::EtsiTs103097Certificate& cert);
```
riebl: Caller might believe `get_issuer_aids` looks up AIDs in this certificate's issuing certificate. This method might be renamed to `get_issue_permissions` in alignment with certificate's field name, what do you think?

vanetza/security/v3/certificate.hpp
```diff
@@ -126,6 +126,8 @@ class CertificateView
      */
     ByteBuffer encode() const;
 
+    const asn1::EtsiTs103097Certificate* raw_certificate() const;
```
riebl: I would like to avoid exposing the internal ASN.1 structure because one of the ideas behind `CertificateView` is to have an interface independent of the used ASN.1 parser.
However, we can add missing getters to `CertificateView` returning "high level" data types (i.e. not types generatey by asn1c).

vanetza/security/v3/certificate.cpp
```diff
@@ -528,6 +533,36 @@ std::list<ItsAid> get_aids(const asn1::EtsiTs103097Certificate& cert)
...
+                    const auto& explicit_perms = group->subjectPermissions.choice.Explicit;
+                    for (int j = 0; j < explicit_perms.list.count; ++j) {
+                        if (explicit_perms.list.array[j]) {
+                            aids.push_back(explicit_perms.list.array[j]->psid);
+                        }
+                    }
+                    break;
+                }
+                case Vanetza_Security_SubjectPermissions_PR_all:
+                    return {};
```
riebl: Current return type makes it impossible to distinguish "no" from "all" permissions. Maybe it is easier to flip the logic: Instead of returning a list of AIDs, add a method like `bool is_allowed_to_issue(ItsAid)`. This would also simplify the calling code.

vanetza/security/v3/certificate_validator.cpp
```diff
@@ -14,6 +19,157 @@ namespace security
...
+    for (int i = 0; i < lhs.list.count; ++i) {
+        if (!lhs.list.array[i] || !rhs.list.array[i] || !equal(*lhs.list.array[i], *rhs.list.array[i])) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool is_within(const asn1::GeographicRegion& inner, const asn1::CircularRegion& outer)
```
riebl: The `is_within` functions are just scratching the surface right now. At minimum, these limitations should be clearly documented in the code.
Ideally, we would have an implementation that can check any `GeographicRegion` if it is within another `GeographicRegion`.

Abderraouf-D:
Thanks for the review @riebl ! I updated the implementation accordingly.

I removed get_issuer_aids() and replaced it with CertificateView::is_allowed_to_issue(ItsAid).

I also removed raw_certificate() from CertificateView and added high-level accessors for the data needed by the validator.
For region containment, I documented the current limitation in CertificateView::region_is_within(). I will try think about a proper generic implementation to do more generic region in region checks.

I also added a regression test for SubjectPermissions::all

vanetza/security/v3/certificate_validator.hpp
```diff
@@ -63,7 +63,9 @@ class DefaultCertificateValidator : public CertificateValidator
 
 private:
     const Certificate* find_issuer_certificate(const CertificateView& certificate) const;
+    bool chain_is_consistent(const CertificateView& signing_cert, ItsAid its_aid) const;
```
riebl, line 69: Since the region checks are not yet fully implemented, I think it would be good to have the option to disable the chain consistency checks and/or the certificate region checks.
Right now, a valid certificate chain might get rejected due to the incomplete implementation of region checks. Methods like `disable_chain_checks` and `disable_region_consistency_checks` would allow an educated user to still run a station in this cases successfully.
What do you think?

Abderraouf-D: Yeah good point, I have just added disable_chain_consistency_checks(bool) and disable_region_consistency_checks(bool) to DefaultCertificateValidator.

vanetza/security/v3/certificate_validator.cpp
```diff
@@ -26,6 +78,8 @@ auto DefaultCertificateValidator::valid_for_signing(const CertificateView& signi
         return Verdict::Expired;
     } else if (!is_chain_anchored(signing_cert)) {
         return Verdict::Untrusted;
+    } else if (!chain_is_consistent(signing_cert, its_aid)) {
+        return Verdict::Unknown;
```
riebl: You have put quite some effort into these consistency checks, so feel free to add `Verdict::InconsistentChain` (or similar) to the enum! :-)

vanetza/security/v3/certificate.cpp
```diff
@@ -144,6 +153,59 @@ bool CertificateView::valid_for_application(ItsAid aid) const
     return m_cert ? v3::valid_for_application(*m_cert, aid) : false;
 }
 
+bool CertificateView::is_allowed_to_issue(ItsAid aid) const
+{
+    if (!m_cert) {
+        return false;
+    }
+
+    if (const auto* seq = m_cert->toBeSigned.certIssuePermissions; seq) {
```
riebl: Isn't the `; seq` superfluous?

vanetza/security/v3/certificate.cpp
```diff
@@ -144,6 +153,59 @@ bool CertificateView::valid_for_application(ItsAid aid) const
...
+                continue;
+            }
+
+            const auto& subject_permissions = group->subjectPermissions;
+            if (subject_permissions.present == Vanetza_Security_SubjectPermissions_PR_all) {
+                return true;
+            } else if (
+                    subject_permissions.present == Vanetza_Security_SubjectPermissions_PR_explicit &&
+                    contains_permission(subject_permissions.choice.Explicit, aid)) {
+                return true;
```
riebl: I think moving the `contains_permission` call to `return contains_permisison(...)` leads to more readable code.

vanetza/security/v3/certificate.cpp
```diff
@@ -144,6 +153,59 @@ bool CertificateView::valid_for_application(ItsAid aid) const
...
+            }
+        }
+    }
+
+    return false;
+}
+
+boost::optional<std::uint8_t> CertificateView::assurance_level() const
+{
+    if (!m_cert || !m_cert->toBeSigned.assuranceLevel || m_cert->toBeSigned.assuranceLevel->size == 0) {
```
riebl: `assuranceLevel` is generated code, so we cannot be absolutely sure that the `size` becomes negative at some point. Thus, I suggest to check `->size < 1` here for more robust code.

riebl (commented):
Looks good, just some minor remarks left. Good work!

vanetza/security/v3/certificate.cpp
Abderraouf-D: I checked the generated type SubjectAssurance (the type of toBeSigned.assuranceLevel )  is currently typedefed to OCTET_STRING_t, and OCTET_STRING_t::size is size_t, so it cannot become negative in the current generated code i guess,
But yes  I agree that checking size < 1 is better, i will change it

vanetza/security/v3/certificate.cpp
Abderraouf-D: I did it to avoid your bot's warnings, the declaration was outside the condition, also I forgot that it can just be if(const auto* seq = m_cert->toBeSigned.certIssuePermissions)

riebl (approved):

riebl:
Ain't I a great reviewer? Just approved your PR and now the CI is failing :laughing:

vanetza/security/v3/certificate.cpp
Abderraouf-D: I tried this simplification, but returning `contains_permission(...)` directly inside the outer loop changed the behavior.

Previously, if one explicit permission group did not contain the requested AID, the loop continued and checked the remaining permission groups. With a direct `return contains_permission(...)`, the function returned `false` immediately on the first non-matching explicit group, so later groups were skipped. This made some valid chains fail

Abderraouf-D:
> Ain't I a great reviewer? Just approved your PR and now the CI is failing 😆

@riebl , aw xD i was just fixing it, i pushed the fixed version in this branch, it was because of the little contains_permission modifications xD Sorry for that

vanetza/security/v3/certificate.cpp
riebl: You are right, sorry for my ill-fated suggestion. At least this has proven that your unit tests are good :-)

